PR #3, opened 2018-08-12
Add dach, a Dai automated clearing house

MrChico:
Adds `dach.sol` and its accompanying test file `dach.t.sol`. 
This is a dss adaptation of the module formerly advertised as [ds-relay](https://github.com/dapphub/ds-relay) which let's users sign `Cheques` which are basic IOUs specifying a dai amount to transfer. A `Cheque` can be cleared by anyone (likely a miner frontrunning anyone else), transferring a dai `fee` to `msg.sender` from the author of the cheque.


asymmetric:
@MrChico can we delete the branch?

PR #10, opened 2018-11-19
Use transfer in exit instead of transferFrom

nanexcool:
Makes `exit` compatible with many ERC20s, not just ds-tokens.


asymmetric:
Because some tokens don’t implement `transferFrom`?

nanexcool:
No, because they way ds-tokens work herehttps://github.com/dapphub/ds-token/blob/master/src/token.sol#L49

Most tokens won't let you do a `transferFrom` from yourself, but ds-tokens do

PR #13, opened 2018-12-11
Make daiMove can access to any urn

gbalabasquer:


asymmetric:
@gbalabasquer could you add an explanation of why this is needed, for my curiosity? thanks :)

levity:
i think this is just correcting a regression

asymmetric:
Ah right, so it's related to the changes in https://github.com/makerdao/dss/commit/88c030d1d5723fdcbfdd728fbbb3bd3522145564

src/move.sol
```diff
@@ -62,7 +62,10 @@ contract DaiMove {
     function hope(address guy) public { can[msg.sender][guy] = 1; }
     function nope(address guy) public { can[msg.sender][guy] = 0; }
     function move(address src, address dst, uint wad) public {
-        require(src == msg.sender || can[src][msg.sender] == 1);
-        vat.move(bytes32(bytes20(src)), bytes32(bytes20(dst)), mul(ONE, wad));
+        move(bytes32(bytes20(src)), bytes32(bytes20(dst)), wad);
```
asymmetric: Does it make sense to do the `bytes20` here? It will be executed in the called function anyway.

gbalabasquer:
Right, I need this now for the new cdp-manager. Otherwise I need to `exit` DAI and then `join` back.
Btw @asymmetric do you know why the CI check fails?

src/move.sol
gbalabasquer: You need to parse to `bytes20` or to `uint256` before parsing to `bytes32`. This is a requirement for solc 0.5. If you parse with `bytes` will align the address to the left, otherwise to the right.

gbalabasquer:
I think it is failing as needs to compile with solc-0.5

rainbreak:
done in cfe4b28, but just deleting the `address` version entirely

PR #18, opened 2019-01-16
DaiMove and GemMove can access to the full rad value in vat

gbalabasquer:


gbalabasquer:
@rainbreak proposed solution to being able to `move` `rad` values of `dai` and `gem`.
Honestly I'm not very convinced this is a good change. For `flip` feels good, but for `flap` and `flop` the difference between managing `dai` in `rad` and `gem` in `wad` doesn't feel right. I guess that is why the `move` objects took care of this conversion.

gbalabasquer:
Closing this PR. I don't think we will `move` the whole `rad` value. In the case we want, we should make a new PR based on the latest changes.

asymmetric:
Should we delete the branch?

gbalabasquer:
Just did.

PR #19, opened 2019-01-18
add messages to requires in frob

levity:
feel free to change the wording to suit you

### DIFF
--- a/src/frob.sol
+++ b/src/frob.sol
@@ -42,7 +42,7 @@ contract Pit is DSNote {
     mapping (address => uint) public wards;
     function rely(address guy) public note auth { wards[guy] = 1; }
     function deny(address guy) public note auth { wards[guy] = 0; }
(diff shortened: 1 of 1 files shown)


gbalabasquer:
@rainbreak I think we should do it now for the whole project. Were you thinking in some type of code styling for them?

levity:
Looking forward to this landing in `vat.sol` now :)

rainbreak:
@levity the reason for delaying here is trying to avoid ossifying prematurely, as there is potentially quite a lot of stuff to add checks for. Do you have an opinion on error codes vs messages? I'm leaning in favour of codes currently.

levity:
I'm totally fine with codes; I just want to be able to tell, when I get a revert, which require is causing it. The first time I put them in myself locally I just called them "frob 1", "frob 2", etc. :)

PR #21, opened 2019-01-18
Debt dust limit. Closes #20

gbalabasquer:


gbalabasquer:
This was done in another PR.

PR #22, opened 2019-01-23
uint multiplication for flip, flap n' flop

MrChico:


xwvvvvwx:
Why?

src/heal.sol
```diff
@@ -71,8 +71,10 @@ contract Vow is DSNote {
         z = x - y;
         require(z <= x);
     }
-    function mul(uint x, uint y) internal pure returns (uint z) {
-        require(y == 0 || (z = x * y) / y == x);
+    function mul(uint x, uint y) internal pure returns (int z) {
```
asymmetric: wut? why does this one get changed to the supposedly uglier implementation?

src/heal.sol
```diff
@@ -116,14 +118,12 @@ contract Vow is DSNote {
     // Debt settlement
     function heal(uint wad) public note {
         require(wad <= Joy() && wad <= Woe());
-        require(int(mul(wad, ONE)) >= 0);
-        VatLike(vat).heal(bytes32(bytes20(address(this))), bytes32(bytes20(address(this))), int(mul(wad, ONE)));
+        VatLike(vat).heal(bytes32(bytes20(address(this))), bytes32(bytes20(address(this))), mul(wad, ONE));
```
asymmetric: Oh so it's about not having the type conversion?

MrChico: Or rather, factoring out the type conversion to the multiplication function. I'm fine with omitting the second commit though.

asymmetric: I slightly prefer the consistent implementation option FWIW.

asymmetric:
Closed because already in https://github.com/makerdao/dss/commit/0e67ddccd73818278f7674880961f2ef9d08e065?

MrChico:
Yea

PR #24, opened 2019-03-18
Math opinions

MrChico:


MrChico:
Negative `add` does not have the same behaviour as `sub`, since `-X` is solidity for `SUB 0 X`, which in the case of `X = 2^255` returns `X`.

PR #26, opened 2019-03-27
request slots for art slab

nmushegian:
Is this approach ok?
>inb4 failing CI below

### DIFF
--- a/src/vat.sol
+++ b/src/vat.sol
@@ -50,6 +50,8 @@ contract Vat {
     mapping (address => uint256)                   public dai;  // [rad]
     mapping (address => uint256)                   public sin;  // [rad]
 
+    bytes32 public slab;  // Maker Art IPFS Hash
+    bytes32 public rest; //   First 32+8 bytes
     uint256 public debt;  // Total Dai Issued    [rad]
     uint256 public vice;  // Total Unbacked Dai  [rad]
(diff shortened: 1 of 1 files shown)


nmushegian:
context:
https://www.reddit.com/r/MakerDAO/comments/b5s0al/mkr_roll_call_vote_in_1_week_serious_vote_fun/
https://www.reddit.com/r/MakerDAO/comments/b66y32/example_result_of_slab_etchings/

src/vat.sol
```diff
@@ -50,6 +50,9 @@ contract Vat {
     mapping (address => uint256)                   public dai;  // [rad]
     mapping (address => uint256)                   public sin;  // [rad]
 
+    bytes32 public slab = 0x0;  // Maker Art IPFS Hash
+    bytes32 public rest = 0x0;  // First 32+8 bytes
```
godsflaw: I see you removed the ability to set these from the constructor, are we just waiting for another commit where you will either hardcode values or add a setter?

nmushegian: Will make another commit when we have the values

BIGCRYPTO001:
Please Sucure This Address
0xb7bBF6D2f5372022cc07B6677A9eB49ce78dA8c3

livnev:
@nmushegian would be more convenient if you added these as the last storage variables declared, since then the storage layout won't need to be remapped for FV.

PR #27, opened 2019-04-07
Dai, dai.t.sol; Use bytes32 instead of dynamic string type.

MrChico:


asymmetric:
So we're not doing this? What was the reason in the first place? ;)

MrChico:
Strings are crazy. But I put some other touchups for the Dai contract here so let's not delete yet

PR #31, opened 2019-05-01
Remove EthJoin from dss; weth suffices

MrChico:
With https://github.com/makerdao/dss-deploy/pull/14, `EthJoin` is no longer a part of MCD deployment


asymmetric:
It's kinda cool to have it for educational/illustrative purposes, and the contracts in `join.sol` are kind of illustrative only, right?

If the maintenance burden to keep it is not too big, I'd keep it.

PR #33, opened 2019-05-23
Pip value is a wad

gbalabasquer:
The actual `end` module doesn't work with the current full implementation as `pip.read` is returning a `wad` instead of a `ray`.
This PR implements the easiest solution IMO.

The other ones would be:
- Changing everything else and making the feeds use rads. I think at this point that will be chaotic.
- Implement the solution deducing the price from the `spot` `mat` and `par` values. Which I think it will be the best way to abstract the precision of the feed itself + the benefit of not needing to whitelist the `end`.

### DIFF
--- a/src/end.sol
+++ b/src/end.sol
@@ -255,7 +255,7 @@ contract End {
         require(live == 0);
         require(tag[ilk] == 0);
         Art[ilk] = vat.ilks(ilk).Art;
-        tag[ilk] = rdiv(RAY, uint(spot.ilks(ilk).pip.read()));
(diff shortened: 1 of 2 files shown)


livnev:
@gbalabasquer good `spot` 👍 
It's slightly simpler and loses less precision to do this though: 6da27caf2f5f36b52178019dcacc60bac5a15be3

gbalabasquer:
@livnev right. So you guys agree with this solution? Can we get it merged?

livnev:
@gbalabasquer done b86991d99aa786aa42bc28afc381b20c908a9b8e

gbalabasquer:
@livnev Thanks!

PR #36, opened 2019-05-26
Merge spotter

desaperados:
Move the Spotter + commit history into dss:
```
github/makerdao/dss-deploy/src/poke.sol -> src.spot.sol
```


rainbreak:
@gbalabasquer I like this. How about you?

gbalabasquer:
@rainbreak same here, like it too.

src/test/spot.t.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity >=0.5.0;
+
+import "ds-test/test.sol";
+
+import {Spotter} from "../spot.sol";
+
+contract SpotterTest is DSTest {
```
rainbreak: Is there any point having this stubbed?

desaperados: I guess it's pretty redundant. Reasoning was that it highlights lack of tests and suggests that we write some, but not a considered choice to include it - happy to remove.

PR #38, opened 2019-05-29
vow, jug, pot: remove unneccessary restrictions on time variables

MrChico:
My humble opinion.

### DIFF
--- a/src/jug.sol
+++ b/src/jug.sol
@@ -22,7 +22,7 @@ contract Jug is DSNote {
     // --- Data ---
     struct Ilk {
         uint256 duty;
-        uint48  rho;
+        uint256  rho;
     }
 
     mapping (bytes32 => Ilk) public ilks;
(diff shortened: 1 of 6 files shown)


asymmetric:
Why had 48 been chosen? 🕰

MrChico:
2^48 = 281474976710656
Is equivalent to:
12/07/8921556 @ 10:44am (UTC)

Maybe during some less ambitious whiteboarding session it was decided that humanity would have transcended debt based interaction patterns at this point

asymmetric:
Should've asked: why not 42?

MrChico:
In auctions, two time variables are packed together with an address, fitting in one storage slot as. 
160+48+48=256
Rest of dss was adapted for consistency. But it introduces some annoying masking and stuff and doesn't really bring any benefit elsewhere, only more gas, bytecode and less aesthetics

PR #47, opened 2019-07-10
Add pit to the vow instead of fixed address(0)

gbalabasquer:


gbalabasquer:
Closing this as `flap` will call directly `gov.burn`

PR #49, opened 2019-08-06
add rho check to join

iamchrissmith:
Protecting against `Pot.join` + `Pot.exit` single transaction profitability.

### DIFF
--- a/src/pot.sol
+++ b/src/pot.sol
@@ -123,7 +123,7 @@ contract Pot is DSNote {
 
     // --- Savings Rate Accumulation ---
     function drip() external note {
-        require(now >= rho);
+        require(now >= rho, "no-double-drip");
         uint chi_ = sub(rmul(rpow(dsr, now - rho, ONE), chi), chi);
         chi = add(chi, chi_);
         rho = now;
@@ -132,6 +132,7 @@ contract Pot is DSNote {
 
     // --- Savings Dai Management ---
     function join(uint wad) external note {
+        require(now == rho, "no-stale-chi");
         pie[msg.sender] = add(pie[msg.sender], wad);
         Pie             = add(Pie,             wad);
         vat.move(msg.sender, address(this), mul(chi, wad));
--- a/src/pot.sol
+++ b/src/pot.sol
@@ -123,7 +123,7 @@ contract Pot is DSNote {
 
     // --- Savings Rate Accumulation ---
     function drip() external note {
-        require(now >= rho, "no-double-drip");
+        require(now >= rho);
         uint chi_ = sub(rmul(rpow(dsr, now - rho, ONE), chi), chi);
         chi = add(chi, chi_);
         rho = now;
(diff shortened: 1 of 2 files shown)


src/pot.sol
```diff
@@ -123,7 +123,7 @@ contract Pot is DSNote {
 
     // --- Savings Rate Accumulation ---
     function drip() external note {
-        require(now >= rho);
+        require(now >= rho, "no-double-drip");
```
rainbreak: You might want to leave the messages empty if this is going to be used as a lemma in FV somewhere. If you still want to keep them then please prefix with `dss/`.

src/pot.sol
```diff
@@ -132,6 +132,7 @@ contract Pot is DSNote {
 
     // --- Savings Dai Management ---
     function join(uint wad) external note {
+        require(now == rho);
```
kmbarry1: It's friendlier to proxies if you change this to:
```
require(now > rho);
if (now == rho) return;
```
Otherwise the proxy needs to compare `rho` to `now` before calling `drip()` when joining.

iamchrissmith: I'm not sure I follow. A proxy would always call `drip` which would succeed if `now >= rho`. Once that succeeds, `rho == now` and we want `join` to proceed if `now == rho` not return.

gbalabasquer: Yes @iamchrissmith is right. The only thing we need to add to the proxy is checking `now >= pot.rho()` before calling `pot.drip` so we avoid a failure due someone else already called `drip` in the same block.

gbalabasquer (approved):
LGTM

src/pot.sol
kmbarry1: I mean, that was my point exactly...if you'd rather do the check in the proxy, go for it. I just don't think it's particularly good interface design (it doesn't really make sense to revert if `drip` already got called within a block...natural behavior is to succeed silently without nuking the transaction).

iamchrissmith: I don't think it reverts if `drip` was already called. `drip` requires it to be greater than or equal, so as long as rho is not somehow set in the future (I suspect this protects against some sort of reentrancy) then `drip` will work. So the proxy should just always call `drip` then `join` if `drip was already called in the block, it will still succeed.

iamchrissmith: i added a test that I believe shows this: https://github.com/makerdao/dss/pull/49/files#diff-d51c5f85148f5ee3cacb45c70fa7680aR79

kmbarry1: Argh, I commented at the wrong place (meant to put this up at 126), and completely misread the direction of the inequality sign. Please ignore.

gbalabasquer: Oh yes, you are right @iamchrissmith . It will not fail.

So @kmbarry1 are you saying to add `if (now == rho) return;` to the the `drip` function? that would actually save gas if it is called twice in a block. But the `require` should still be `>=` or be after the `if/return` statement.

Anyway we should consider if saving gas for these double cases is more convenient that spending a bit more of gas in every call just checking.

kmbarry1: Yeah, what I was trying to propose amounts to a gas optimization if there are multiple `Pot.drip()` calls per block. I originally had a confusion that `drip()` would revert if `now == rho` (the justification for which is lost on me...bit flip somewhere in my brain). I actually doubt this optimization alone does much, as `rpow` terminates very quickly when the power input is zero.

PR #50, opened 2019-08-07
Add tick to flop

iamchrissmith:
Adding a tick to flop. This allows for restarting the Flop auction if it reaches the `end` before receiving any `bid`s.

### DIFF
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -109,6 +109,11 @@ contract Flopper is DSNote {
 
         emit Kick(id, lot, bid, gal);
     }
+    function tick(uint id) external note {
+        require(bids[id].end < now);
+        require(bids[id].tic == 0);
+        bids[id].end = add(uint48(now), tau);
+    }
     function dent(uint id, uint lot, uint bid) external note {
         require(live == 1);
         require(bids[id].guy != address(0));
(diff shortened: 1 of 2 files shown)


andy8052:
Having `tic` and `tick` could be a little confusing, but I also can't think of another name for the function. Meme function & variable names have never been a forte of mine

iamchrissmith:
This just mirrors the `tick` function in `Flip`, so I think it is best to be consistent (and I certainly don't have better ideas on the memes either :-) )

andy8052 (approved):

gbalabasquer:
LGTM

MrChico:
If one changes the size of the lot, from the slightly overzealous `uint(-1)` to, say a variable subject to `filing`, then the `tick` function could also bump up the lot at restart

iamchrissmith:
> If one changes the size of the lot, from the slightly overzealous `uint(-1)` to, say a variable subject to `filing`, then the `tick` function could also bump up the lot at restart

So instead of starting an auction with `uint(-1)` on offer, start it with X MKR and then if no one bids, the system ups its offer to X * `beg` (for instance) and restarts the auction to see if people now want to bid?

MrChico:
Essentially, yeah

iamchrissmith:
@MrChico that makes sense, though I think it is too big for this PR.  I created another branch that addresses that potential change: https://github.com/makerdao/dss/pull/52
Lets discuss the impact/implementation there

PR #51, opened 2019-08-07
Add move function to pot

gbalabasquer:


rainbreak:
There's another way of thinking about this which is to have the user `vat.hope` some contract, which would then have the ability to take the user's dai and save it, while exposing some token interface to the user.

gbalabasquer:
@rainbreak right. But I still think this could give some more flexibility in the future to implement a more transparent ERC20 SDAI.
Do you thing it could harm anyway to have these extra functions here?

PR #52, opened 2019-08-07
Flop varying sized mkr lots

iamchrissmith:
Currently Vow uses `uint(-1)` as the starting point for Flop auctions.  This is:
1) a huge but arbitrary number
2) will cause the auction to fail if there are no bids as it is not possible to mint uint(-1) new MKR because `totalSupply` would overflow
3) it opens the possibility of low bidder participation minting a huge amount of MKR.

This introduces `Vow.dump` which would be the governance-controlled variable for how much `lot` (i.e. MKR gems) the Flop auction should start with. The `dump` replaces `uint(-1)` as the ceiling from which to start the `Flop`/`dent` auctions. If bidders do not feel this amount is worth the asking `bid` and the auction expires with 0 bids, then `tick` will increase the `lot` by `beg` and restart the auction.

Benefits of this approach:
- Since it is governance set, it is no longer arbitrary
- As long as Governance acts responsibly:
 - this number should not be so large to cause an overflow
 - this number should not represent a "thread" to MKR governance in the case of low bidder participation

However, this approach:
- requires an additional governance variable, increasing complexity
- needs further discussion around the economic/game theory consequences as it possibly introduces an incentive for bidders to wait until `tick` can be called.

(updated to change `pump` to `dump`)

### DIFF
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -60,6 +60,7 @@ contract Flopper is DSNote {
 
     uint256  constant ONE = 1.00E27;
     uint256  public   beg = 1.05E27;  // 5% minimum bid increase
+    uint256  public   leg = 1.50E27;  // 50% lot increase for tick
     uint48   public   ttl = 3 hours;  // 3 hours bid lifetime
     uint48   public   tau = 2 days;   // 2 days total auction length
     uint256  public kicks = 0;
@@ -92,6 +93,7 @@ contract Flopper is DSNote {
     // --- Admin ---
     function file(bytes32 what, uint data) external note auth {
         if (what == "beg") beg = data;
+        else if (what == "leg") leg = data;
         else if (what == "ttl") ttl = uint48(data);
         else if (what == "tau") tau = uint48(data);
         else revert();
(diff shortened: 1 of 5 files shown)


MrChico:
I would not reuse `beg` here, but use a new meme. My gut feeling is that the restart factor should be around 2, not ~1.1

rainbreak:
Perhaps `dump` is more apt than `pump` :)

iamchrissmith:
> I would not reuse `beg` here, but use a new meme. My gut feeling is that the restart factor should be around 2, not ~1.1

I debated that and figured since we were reusing beg for both `tend` and `dent` in Flip auctions, I could do the same here. However, I can definitely agree that having a separate increase factor could be important.  I'm not sure I would agree it should be double, so I set the default (still arbitrarily) to 1.5x. 

Meme-wise: I'm still trying to strengthen those creative muscles.  I could not come up with a great candidate from a topical sense so I took the `dart`/`dink` path and called it `leg` (`beg` for `bid` and `leg` for `lot`).  I'm hoping someone else has a better idea :-)

levity:
> Meme-wise: I'm still trying to strengthen those creative muscles. I could not come up with a great candidate from a topical sense so I took the `dart`/`dink` path and called it `leg` (`beg` for `bid` and `leg` for `lot`). I'm hoping someone else has a better idea :-)

`pad`?

levity:
My first contribution to the meme language 😁

PR #53, opened 2019-08-16
add end to the pot to prevent drips after global settlement starts

iamchrissmith:
Add Cage to the Pot so that when we start the `End` `Pot.drip` will no longer accumulate `DAI`

### DIFF
--- a/src/end.sol
+++ b/src/end.sol
@@ -53,6 +53,9 @@ contract CatLike {
     function ilks(bytes32) external returns (Ilk memory);
     function cage() external;
 }
+contract PotLike {
+    function cage() external;
+}
 contract VowLike {
     function heal(uint256 rad) external;
     function cage() external;
@@ -94,6 +97,7 @@ contract Spotty {
         - freezes user entrypoints
         - cancels flop/flap auctions
         - starts cooldown period
+        - stops pot drips
 
     2. `cage(ilk)`:
        - set the cage price for each `ilk`, reading off the price feed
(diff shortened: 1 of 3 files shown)


gbalabasquer (approved):
LGTM

kmbarry1:
LGTM. I guess we should update the spec for `Pot.drip` in k-dss too.

kmbarry1:
I rather like Lev's point about setting `dsr = RAY` (might break a bunch of dapps and even prevent Dai redemption)...although this also requires adding a `live` check to `file` so that malicious governance is prevented from changing the DSR value post-`cage`.

gbalabasquer:
@kmbarry1 malicious governance can replace directly the `pot` contract by a new one, excepting we put the `require(live == 1)` in the `rely` function of the `vat`.

kmbarry1:
The Pauses for those two actions will be very different though--DSR will be O(days) for economic responsiveness, `rely`/`deny` against the Vat will be O(months). Given that ES will take longer than a few days, in all likelihood, I think it's meaningful.

Also your argument that malicious governance can replace the Pot is irrelevant to choosing between the two solutions because they can do in the "disable `drip`" solution just as easily.

gbalabasquer:
Yeah that is clear, it doesn't matter the type of protection we are setting in the `pot` if it can be eventually replaced. So I was not really meaning any difference between options regarding this problem.
However it is true that will be less problem with a longer pause for the `vat.rely/vat.deny` functions.

kmbarry1 (approved):
No objections here.

PR #54, opened 2019-08-21
correct inconsistent time type in vow

iamchrissmith:
Makes `wait` a `uint48` which matches the type used for other time variables.


livnev:
For reference: `uint48` is used for times in the auction data structures to take advantage of storage packing, since bidding uses a lot of storage. In the `vow` it doesn't make any practical difference, other than potentially introducing some theoretical masking annoyances in FV.

PR #57, opened 2019-09-04
update cage(ilk) to account for the dai par

iamchrissmith:
Currently, the `Spot.par` or `ref per dai` is not accounted for during the `End`. This updates `cage(ilk)` to :
1 - Call `spot.poke(ilk)` to get the updated price in the `vat`
2 - Retrieves that `spot` price and the `spot.ilks(ilk).mat`
3 - multiplies the `spot` by the `mat` to calculate the ilk's tag with the `par` accounted for.

thank you to @gbalabasquer for suggesting this approach as a way of pulling in the `par`. 

As he points out we should evaluate if this is the best approach specifically in relation to its impact on FV for the `End`.

### DIFF
--- a/src/end.sol
+++ b/src/end.sol
@@ -84,6 +84,7 @@ contract Spotty {
         PipLike pip;
         uint256 mat;
     }
+    function poke(bytes32) external;
     function ilks(bytes32) external view returns (Ilk memory);
 }
 
@@ -270,8 +271,9 @@ contract End is DSNote {
         require(live == 0);
         require(tag[ilk] == 0);
         Art[ilk] = vat.ilks(ilk).Art;
-        // pip returns a wad, invert it and convert to ray
-        tag[ilk] = rdiv(WAD, uint(spot.ilks(ilk).pip.read()));
+        // spot and mat are both RAY, so multiple then invert it
+        spot.poke(ilk);
+        tag[ilk] = rdiv(RAY, rmul(vat.ilks(ilk).spot, spot.ilks(ilk).mat));
     }
 
     function skip(bytes32 ilk, uint256 id) external note {
--- a/src/test/end.t.sol
+++ b/src/test/end.t.sol
@@ -37,26 +37,6 @@ contract Hevm {
     function warp(uint256) public;
 }
 
-// contract TestSpot {
-//     struct Ilk {
-//         address pip;
-//         uint256 mat;
-//     }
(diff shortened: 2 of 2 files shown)


gbalabasquer (approved):
The change looks good to me.

livnev:
@iamchrissmith This is not what you want, since `mat` is the liquidation ratio. I think you want something like #59.

iamchrissmith:
> @iamchrissmith This is not what you want, since `mat` is the liquidation ratio. I think you want something like #59.

Yeah I think that is another way to achieve it.  However, in calling `spot.poke(ilk)` we are setting the `vat.ilks(ilk).spot` to `= (val / par) / mat`, so doesn't this approach achieve the same thing by multiplying `spot` by `mat` (i.e. `((val / par) / mat ) * mat == val / par`

iamchrissmith:
If that is the case, I think then the decision comes down to whether it is better to call poke in the `End` or just pull the `read` value. Do you remember why you all decided to use `pip.read` instead of `spot.poke`? (I believe @gbalabasquer said this required whitelisting the `End` in the `pip`s but maybe it was because that approach is easier for FV?)

livnev:
> > @iamchrissmith This is not what you want, since `mat` is the liquidation ratio. I think you want something like #59.
> 
> Yeah I think that is another way to achieve it. However, in calling `spot.poke(ilk)` we are setting the `vat.ilks(ilk).spot` to `= (val / par) / mat`, so doesn't this approach achieve the same thing by multiplying `spot` by `mat` (i.e. `((val / par) / mat ) * mat == val / par`

Oh yes, of course, I somehow missed this. So basically your way effectively calculates `tag` as:

```
rdiv(RAY, rmul(rdiv(rdiv(mul(pip, 10 ** 9), par), mat), mat))
```

while my way was

```
wdiv(par, pip)
```

Note that the former introduces two additional sources of rounding error whenever `par != RAY` or `mat != RAY`, since it's not the case that `rdiv(a, b) * b == b`.

iamchrissmith:
@gbalabasquer I know you said one benefit of using this approach was eliminating the whitelist dependency.  However, at this point I feel it introduces new contract interaction/dependency and as @livnev points out more rounding issues.  I would favor going with #59 over this approach. What are your thoughts?

gbalabasquer:
Yeah, let's go with https://github.com/makerdao/dss/pull/59

PR #59, opened 2019-09-06
account for spot.par in the End

livnev:


iamchrissmith:
Info about why:
Currently, the Spot.par or ref per dai is not accounted for during the End. This updates cage(ilk) to account for `par`

PR #60, opened 2019-09-08
revert `file` calls if `what` is unrecognized

kmbarry1:
Addresses TOB-MCD-011. Will update k-dss as well.

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -100,13 +100,16 @@ contract Cat is DSNote {
     // --- Administration ---
     function file(bytes32 what, address data) external note auth {
         if (what == "vow") vow = VowLike(data);
+        else revert();
     }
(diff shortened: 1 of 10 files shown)


iamchrissmith:
I would have a slight preference to handle that separately because then we can point to separate PRs in response to the different issues.

gbalabasquer:
Sounds good.

kmbarry1:
@gbalabasquer yes, separate PR was my plan, for the exact reason that Chris mentions.

asymmetric:
Is the issue public?

kmbarry1:
@asymmetric  it will be public once the audit reports are released, but I can summarize quickly: this is just to prevent silent failures if one mistypes the `what` argument, not to mitigate any real vulnerability. This avoids issues like thinking a parameter has been changed when in fact it has not, and can speed up debugging cycles.

PR #63, opened 2019-09-17
Minor code clean-up for Vat.fork

iamchrissmith:
One of the auditors noticed inconsistent usage of `&&` and `||` in `requires` for `vat.fork`. This just switches them to the frob-like `either` and `both`.  I do not believe this requires changes to FV, but let me know if it does and I'll close this.

### DIFF
--- a/src/vat.sol
+++ b/src/vat.sol
@@ -210,15 +210,15 @@ contract Vat {
         uint vtab = mul(v.art, i.rate);
 
         // both sides consent
-        require(wish(src, msg.sender) && wish(dst, msg.sender));
+        require(both(wish(src, msg.sender), wish(dst, msg.sender)));
 
         // both sides safe
         require(utab <= mul(u.ink, i.spot));
         require(vtab <= mul(v.ink, i.spot));
 
         // both sides non-dusty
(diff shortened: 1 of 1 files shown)


kmbarry1 (approved):
LGTM

livnev:
You may want to tests `k-dss` with this change in case any changes are needed to lemmas etc.

gbalabasquer (approved):
I was thinking exactly about what @livnev says. I'd sugges we wait for merging this change until we get our `k-dss` set up working, then we just test this specific change and see if something breaks on the FV side. Btw the PR looks good.

iamchrissmith:
Sounds good, I'll leave this until we can test in `k-dss`

MrChico:
This should be easier from an fv perspective than the current master, but yeah, to be sure one should rerun k-dss

gbalabasquer:
@kmbarry1 taking into account what we discuss today, should we just merge this now? In theory it shouldn't bring any FV issue.

kmbarry1:
Yeah might as well merge...might need some changes to the lemmas but given the backlog we already have, I think it's fine.

PR #64, opened 2019-09-19
more gas efficient implementation of pot.drip

manuelpolzhofer:
hey,

I noticed in the `pot.drip` method the addition `chi = add(chi, chi_);` is more expensive than just to store the latest chi (calculated already in the previous line).

Improvement is around `~600 gas` per method call.

Since the method should be called very often. I think it makes sense to change it.

**Current Gas**
```
Running 7 tests for src/test/pot.t.sol:DSRTest
[PASS] testFail_stale_chi() (gas: 145165)
[PASS] test_drip_multi() (gas: 549591)
[PASS] test_drip_multi_inBlock() (gas: 375922)
[PASS] test_fresh_chi() (gas: 457704)
[PASS] test_save_0d() (gas: 454693)
[PASS] test_save_1d() (gas: 522272)
[PASS] test_save_multi() (gas: 761726)
```

**Improvement Gas**
```
Running 7 tests for src/test/pot.t.sol:DSRTest
[PASS] testFail_stale_chi() (gas: 144631)
[PASS] test_drip_multi() (gas: 548523)
[PASS] test_drip_multi_inBlock() (gas: 374320)
[PASS] test_fresh_chi() (gas: 457170)
[PASS] test_save_0d() (gas: 454159)
[PASS] test_save_1d() (gas: 521738)
[PASS] test_save_multi() (gas: 760658)
```

### DIFF
--- a/src/pot.sol
+++ b/src/pot.sol
@@ -135,8 +135,9 @@ contract Pot is DSNote {
     // --- Savings Rate Accumulation ---
     function drip() external note {
         require(now >= rho);
-        uint chi_ = sub(rmul(rpow(dsr, now - rho, ONE), chi), chi);
-        chi = add(chi, chi_);
+        uint latest = rmul(rpow(dsr, now - rho, ONE), chi);
+        uint chi_ = sub(latest, chi);
+        chi = latest;
         rho = now;
         vat.suck(address(vow), address(this), mul(Pie, chi_));
     }
--- a/src/pot.sol
+++ b/src/pot.sol
@@ -135,9 +135,9 @@ contract Pot is DSNote {
     // --- Savings Rate Accumulation ---
     function drip() external note {
         require(now >= rho);
-        uint latest = rmul(rpow(dsr, now - rho, ONE), chi);
-        uint chi_ = sub(latest, chi);
-        chi = latest;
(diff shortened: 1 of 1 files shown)


andy8052:
Other than potentially changing `latest` to be a variable name that fits with the naming scheme of `dss` I don't see a reason to not get this merged. Interested to hear others thoughts though.

kmbarry1:
I'm in favor of merging as well. Of course gas costs can totally change with the next hard fork, but I don't think there's any downside to this PR (okay, an extra line of code and a temp variable, but that's negligible IMO).

Re: naming...I don't immediately have a good idea for a shorter name than `latest` that's equally descriptive. I guess we could just use `tmp` or something similarly generic, but OTOH I also don't really see the harm in having a slightly-longer-named variable here.

manuelpolzhofer:
let me know if I should update the variable name.

andy8052:
Considering we already have a 3 and 4 length variable name, I am fine to just leave it

gbalabasquer:
I'd say we merge this, but if you agree I'll just rename it to `tmp` or `aux`.

manuelpolzhofer:
I changed it to `tmp`

PR #66, opened 2019-09-20
flap/flip/flop.sol: beg and pad are wad now

gbalabasquer:
I was also thinking on change this line in `flop.sol`:

```
require(mul(beg, lot) / ONE <= bids[id].lot);  // div as lot can be huge
```
to:
```
require(mul(beg, lot) <= mul(bids[id].lot, ONE));
```

As now the initial `bids[id].lot` is more controlled and `ONE` is smaller. But let me know what you think about it.

### DIFF
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -58,8 +58,8 @@ contract Flapper is DSNote {
     VatLike  public   vat;
     GemLike  public   gem;
 
-    uint256  constant ONE = 1.00E27;
-    uint256  public   beg = 1.05E27;  // 5% minimum bid increase
+    uint256  constant ONE = 1.00E18;
+    uint256  public   beg = 1.05E18;  // 5% minimum bid increase
     uint48   public   ttl = 3 hours;  // 3 hours bid duration
     uint48   public   tau = 2 days;   // 2 days total auction length
     uint256  public kicks = 0;
--- a/src/flip.sol
+++ b/src/flip.sol
@@ -62,8 +62,8 @@ contract Flipper is DSNote {
     VatLike public   vat;
     bytes32 public   ilk;
 
-    uint256 constant ONE = 1.00E27;
-    uint256 public   beg = 1.05E27;  // 5% minimum bid increase
+    uint256 constant ONE = 1.00E18;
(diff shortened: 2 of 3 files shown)


iamchrissmith (approved):
lgtm

iamchrissmith:
I would definitely change that comment, it never felt all that helpful to me in the first place.

gbalabasquer:
The reality is if we change the division by a `mul` in the other side, we will be reducing by 18 digits the initial maximum `lot` size. However with the actual solution (leaving the math as it is) the initial `lot` size could be able to be 18 digits bigger, but the first `bid` will still need to decrease drastically those 18 digits.

Anyway looking at the `tick` function we can notice that is not possible to have an initial `lot` value which will then multiplied by ONE will overflow, as `mul(pad, bids[id].lot)` will previously fail:
```
function tick(uint id) external note {
    require(bids[id].end < now);
    require(bids[id].tic == 0);
    bids[id].lot = mul(pad, bids[id].lot) / ONE;
    bids[id].end = add(uint48(now), tau);
}
```
If we really want to do it correctly, we should verify the maximum possible in `tick` and not allow to increase over that.

So my proposed changes would be in `flop.dent`:
```
require(mul(beg, lot) / ONE <= bids[id].lot);
```
to:
```
require(mul(beg, lot) <= mul(bids[id].lot, ONE));
```

And some check in `flop.tick` that if `mul(pad, bids[id].lot)` overflows, then the `lot` size will be the maximum.

gbalabasquer:
Just added the change in `flop.dent`, let's discuss if we want to add a check in `tick`

gbalabasquer:
To protect against a theoretically (but not practically possible IMO) overflow in `tick`, which would make the auction to be stuck, we would need something like this for that function:

```
function tick(uint id) external note {
        require(bids[id].end < now);
        require(bids[id].tic == 0);
        uint lot = pad * bids[id].lot;
        bids[id].lot = lot > bids[id].lot ? lot : uint(-1) / ONE;
        bids[id].end = add(uint48(now), tau);
    }
```

But not sure if worth it to make this code more complex.

kmbarry1 (approved):
LGTM

PR #67, opened 2019-09-20
add auth to Flip.kick

iamchrissmith:
Current `Flip.kick` is a public function.  It is preferable to only allow the `Cat` to call `kick` to preserve the integrity of auctions

### DIFF
--- a/src/test/flip.t.sol
+++ b/src/test/flip.t.sol
@@ -99,6 +99,7 @@ contract FlipTest is DSTest {
         vat.set_ilk("gems");
 
         flip = new Flipper(address(vat), "gems");
+        flip.rely(address(this));
 
         ali = address(new Guy(flip));
         bob = address(new Guy(flip));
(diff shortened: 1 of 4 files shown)


gbalabasquer (approved):
LGTM

src/test/flip.t.sol
```diff
@@ -99,6 +99,7 @@ contract FlipTest is DSTest {
         vat.set_ilk("gems");
 
         flip = new Flipper(address(vat), "gems");
+        flip.rely(address(this));
```
gbalabasquer: This is actually not necessary as the `flipper` is created by `this`

kmbarry1:
LGTM. This requires an update to the deployment logic too, I assume.

kmbarry1 (approved):

gbalabasquer:
@kmbarry1 yes, already have this issue assigned: https://github.com/makerdao/smart-contracts-team-pm/issues/337 to work on that

PR #68, opened 2019-09-20
add auth to Flap.kick so that only Vow can call it

iamchrissmith:
Current Flip.kick is a public function. It is preferable to only allow the Vow to call kick to preserve the integrity of auctions

### DIFF
--- a/src/test/end.t.sol
+++ b/src/test/end.t.sol
@@ -191,7 +191,6 @@ contract EndTest is DSTest {
         gov.setOwner(address(flop));
 
         vow = new Vow(address(vat), address(flap), address(flop));
-        flap.rely(address(vow));
 
         cat = new Cat(address(vat));
         cat.file("vow", address(vow));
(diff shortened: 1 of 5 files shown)


src/test/end.t.sol
```diff
@@ -191,6 +191,7 @@ contract EndTest is DSTest {
         gov.setOwner(address(flop));
 
         vow = new Vow(address(vat), address(flap), address(flop));
+        flap.rely(address(vow));
```
gbalabasquer: This can be removed, this same permission was already set at the end of the `setUp` function.

src/test/flap.t.sol
```diff
@@ -56,6 +56,7 @@ contract FlapTest is DSTest {
         gem = new DSToken('');
 
         fuss = new Flapper(address(vat), address(gem));
+        fuss.rely(address(this));
```
gbalabasquer: Same here, `fuss` is created by `this`.

gbalabasquer (approved):
LGTM, just these two `rely` that can be removed.

kmbarry1 (approved):

PR #72, opened 2019-09-25
ISSUE-335-02: remove = operator as it's the default

godsflaw:
Since the `=` operator is assumed if just given the pragma, we will just remove it.  The only change here is to make the pragma line more sexy.


andy8052:
Nothing like a sexy pragma line

PR #73, opened 2019-10-01
Set spot to 0 for ilk on hold

gbalabasquer:


src/spot.sol
```diff
@@ -84,10 +84,8 @@ contract Spotter is DSNote {
     // --- Update value ---
     function poke(bytes32 ilk) external {
         (bytes32 val, bool zzz) = ilks[ilk].pip.peek();
-        if (zzz) {
-            uint256 spot = rdiv(rdiv(mul(uint(val), 10 ** 9), par), ilks[ilk].mat);
-            vat.file(ilk, "spot", spot);
-            emit Poke(ilk, val, spot);
-        }
+        uint256 spot = zzz ? rdiv(rdiv(mul(uint(val), 10 ** 9), par), ilks[ilk].mat) : 0;
```
iamchrissmith: Why did you remove the `zzz` check? In the case of intentionally setting `spot` to `0` would we not leave the `zzz` to `true` to signal it was different than just an uninitialized value?

gbalabasquer: If the osm is voided: https://github.com/makerdao/osm/blob/master/src/osm.sol#L73 or expired.
Then you will receive `zzz = false`. In that case we would want the `ilk` is "on hold". The way we propose to do that is setting `spot` to zero.

iamchrissmith: That makes sense, I guess I thinking it might be helpful to be able to distinguish between an osm that is voided and one that is just not setup (if I understand this correctly, now they will both return `(0, false)`).  Probably doesn't matter, so lgtm

iamchrissmith (approved):

src/spot.sol
gbalabasquer: Yeah I guess you will not be able to distinguish that easily, at least if you are not seeing the history of it.

iamchrissmith (approved):

src/spot.sol
```diff
@@ -84,10 +84,8 @@ contract Spotter is DSNote {
     // --- Update value ---
     function poke(bytes32 ilk) external {
         (bytes32 val, bool zzz) = ilks[ilk].pip.peek();
```
kmbarry1: Could we change `zzz` to `has` if we're touching this function anyway? `zzz` usually denotes a time interval, not a boolean flag.

kmbarry1:
So, IIUC, the intent of this change is to disable all liquidations if an OSM gets voided?

Just confirming, as neither the description nor the comments make this obvious. We should really try to make the purpose of each commit clear as an easy form of documentation.

I'm conflicted on setting the Vat's spot price for a given ilk to zero to indicate this. If a frozen feed means frozen liquidations (i.e. historical prices don't matter), we arguably shouldn't even *have* a spot price in the Vat, and the Cat should just directly read from the OSM.

Off-topic: have we thought about what happens if ES gets triggered while an OSM is voided?

src/spot.sol
gbalabasquer: Done!

gbalabasquer:
> So, IIUC, the intent of this change is to disable all liquidations if an OSM gets voided?

Yes

> We should really try to make the purpose of each commit clear as an easy form of documentation.

You are right, I should improve the message on that commit.

> I'm conflicted on setting the Vat's spot price for a given ilk to zero to indicate this. If a frozen feed means frozen liquidations (i.e. historical prices don't matter), we arguably shouldn't even _have_ a spot price in the Vat, and the Cat should just directly read from the OSM.

I understand what you mean but I think the main reason to have a `spot` value was FV. I believe it was easier (at least in the initial stages of development). It would make sense to use directly the feed.

> Off-topic: have we thought about what happens if ES gets triggered while an OSM is voided?
This is a really good question. I guess governance should have to fix the feed before being able to call `cage(ilk)` but not sure how the delay will play in that case.

gbalabasquer:
Any other comment on this? Otherwise I'll merge it today.

kmbarry1 (approved):

kmbarry1:
I'd still like to think more about maybe eliminating `spot` in the Vat, but let's not do that in this PR.

PR #75, opened 2019-10-18
Avoid keccak256 in constant

k06a:
Solidity 0.5.x still compiles constants as pure functions, which executes on every access

### DIFF
--- a/src/dai.sol
+++ b/src/dai.sol
@@ -48,7 +48,7 @@ contract Dai is DSNote {
 
     // --- EIP712 niceties ---
     bytes32 public DOMAIN_SEPARATOR;
-    bytes32 public PERMIT_TYPEHASH;
+    bytes32 public constant PERMIT_TYPEHASH = 0xea2aa0a1be11a07ed86d755c93467f4f82362b452371d1ba94d1715123511acb;
 
     constructor(uint256 chainId_) public {
         wards[msg.sender] = 1;
@@ -59,9 +59,6 @@ contract Dai is DSNote {
             chainId_,
             address(this)
         ));
-        PERMIT_TYPEHASH = keccak256(
-            "Permit(address holder,address spender,uint256 nonce,uint256 expiry,bool allowed)"
-        );
     }
 
     // --- Token ---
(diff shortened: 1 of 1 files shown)


MrChico:
If this is the case I would rather hardcore the result of the hash

MrChico:
But I'm not sure this is true. The documentation suggests otherwise. Did you step this through in a debugger?

k06a:
@MrChico this feature is planned for Solidity 0.6.0: https://github.com/ethereum/solidity/issues/3157

OpenZeppelin follows this way:
- [DeprecatedAdminUpgradeabilityProxy.sol:26](https://github.com/OpenZeppelin/openzeppelin-sdk/blob/ef03e13516855757b9bcfda6f0a8aab5534240db/packages/lib/contracts/mocks/DeprecatedAdminUpgradeabilityProxy.sol#L26):

    ```Solidity
    bytes32 private constant ADMIN_SLOT = 0x10d6a54a4754c8869d6886b5f5d7fbfa5b4522237ea5c60d11bc4e7a1ff9390b;
    ```
- [DeprecatedAdminUpgradeabilityProxy.sol:51](https://github.com/OpenZeppelin/openzeppelin-sdk/blob/ef03e13516855757b9bcfda6f0a8aab5534240db/packages/lib/contracts/mocks/DeprecatedAdminUpgradeabilityProxy.sol#L51):

    ```Solidity
    constructor(...) {
        assert(ADMIN_SLOT == keccak256("org.zeppelinos.proxy.admin"));

        // ...
    }
    ```

k06a:
@MrChico replaced with compile-time constant 👌

MrChico:
You're right. Please add a comment with the hash preimage so this doesn't just look like a magic number

k06a:
Added comment

livnev:
Will anyone actually ever call the `PERMIT_TYPEHASH()` getter on-chain? I assumed this would only be used by client/wallet software which will just do an off-chain call, in which the gas consumption of this method is irrelevant?

k06a:
@livnev I mean even access here https://github.com/makerdao/dss/blob/master/src/dai.sol#L123 leads to `keccak256` recalculation.

livnev:
> @livnev I mean even access here https://github.com/makerdao/dss/blob/master/src/dai.sol#L123 leads to `keccak256` recalculation.

Ah yes, that's a good point.

PR #77, opened 2019-10-19
Fix build error by adding missing dependency ds-note

k06a:
This will make commands `make build` and `make test` work directly.


MrChico:
ds-note is no longer being used in this repo. Better to fix the makefile

gbalabasquer:
Makefile fixed. Closing this PR.

PR #78, opened 2019-10-19
Optimize rpow(x,n,base)

k06a:
1. Optimize rpow(x,n,base) for the case x != 0 && n == 0 (291 gas vs 383 gas)

    ```Solidity
    switch x case 0 {switch n case 0 {z := b} default {z := 0}}
    default {
       ...
    }
    ```
    =>
    ```Solidity
    switch n case 0 { z := b }
    default {
        switch x case 0 { z := 0 }
        default {
            ...
        }
    }
    ```

2. Simplify overflow check for squaring in rpow(x,n,base) (2355 gas vs 2388 gas for x = 2, n = 255)

    ```Solidity
    let xx := mul(x, x)
    if iszero(eq(div(xx, x), x)) { revert(0,0) }
    ```
    =>
    ```Solidity
    let xx := mul(x, x)
    if shr(128, x) { revert(0,0) }
    ```

### DIFF
--- a/src/jug.sol
+++ b/src/jug.sol
@@ -38,27 +38,30 @@ contract Jug is DSNote {
 
     // --- Math ---
     function rpow(uint x, uint n, uint b) internal pure returns (uint z) {
-      assembly {
-        switch x case 0 {switch n case 0 {z := b} default {z := 0}}
-        default {
-          switch mod(n, 2) case 0 { z := b } default { z := x }
-          let half := div(b, 2)  // for rounding.
-          for { n := div(n, 2) } n { n := div(n,2) } {
-            let xx := mul(x, x)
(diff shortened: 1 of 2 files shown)


livnev:
@kmbarry1 can you get these optimisations to pass the spec? 😛

kmbarry1:
I will give it a try today!

kmbarry1:
Also, just to give "official" comment on this: looks good, but there are two things I want to check prior to merging:
1) gas for `x != 0 && n != 0` is not worsened (I don't think it is, looks like it should be at least the same, maybe better, but want to check explicitly)
2) must successfully formally verify this

gbalabasquer:
@kmbarry1 do we have any update on this regarding FV?

kmbarry1:
Am getting through the backlog of changes that are definitely going in before analyzing this.

PR #80, opened 2019-10-22
Remove pragma experimental ABIEncoderV2

gbalabasquer:


src/cat.sol
```diff
@@ -114,28 +110,27 @@ contract Cat is DSNote {
...
+        (, uint rate, uint spot) = vat.ilks(ilk);
+        (uint ink, uint art) = vat.urns(ilk, urn);
 
         require(live == 1);
-        require(mul(u.ink, i.spot) < mul(u.art, i.rate));
+        require(mul(ink, spot) < mul(art, rate));
 
-        uint lot = min(u.ink, ilks[ilk].lump);
-        uint art = min(u.art, mul(lot, u.art) / u.ink);
-        uint tab = mul(art, i.rate);
```
MrChico: Why did `tab` go here? Solidity complaining?

gbalabasquer: Yeah `stack too deep`.

PR #82, opened 2019-10-22
add dec field to GemJoin

kmbarry1:
This is preparation to use the `GemJoin` in `dss` as the "official" ERC20 adapter.

### DIFF
--- a/src/join.sol
+++ b/src/join.sol
@@ -20,6 +20,7 @@ pragma solidity 0.5.11;
 import "./lib.sol";
 
 contract GemLike {
+    function decimals() public view returns (uint);
     function transfer(address,uint) external returns (bool);
     function transferFrom(address,address,uint) external returns (bool);
 }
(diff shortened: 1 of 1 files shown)


iamchrissmith (approved):

gbalabasquer (approved):

iamchrissmith (approved):

src/join.sol
```diff
@@ -70,7 +70,7 @@ contract GemJoin is DSNote {
     bytes32 public ilk;
     GemLike public gem;
     uint    public dec;
-    uint256 public live;  // Access Flag
+    uint    public live;  // Access Flag
```
MrChico: better to make both variables `uint256` for those extra code alignment points

kmbarry1: It is tempting but then I'd feel obligated to suffixify with 256 throughout the file...

gbalabasquer: Yeah alignment with other words like `address/mapping/GemLIke/VatLike/etc` and explicitness lead me to use `uint256` instead of `uint`. I guess we can have a discussion later and create a PR to standardise it properly in the whole repo.

kmbarry1: Yeah doing it piecemeal is inefficient. At some point we should actually establish strict and explicit coding conventions but I don't view that as urgent.

PR #83, opened 2019-10-24
harden system against interference from governance during global settlement

kmbarry1:
Addresses https://github.com/makerdao/smart-contracts-team-pm/issues/331

I tried to make the minimal changes necessary while also erring on the side of caution (so I didn't lock down, for example, `file` calls that wouldn't affect the progression of GS).

Notable breaking change: Spot now has a `cage()` method.

### DIFF
--- a/src/pot.sol
+++ b/src/pot.sol
@@ -129,7 +129,6 @@ contract Pot is DSNote {
     }
 
     function cage() external note auth {
-        require(live == 1);
         live = 0;
         dsr = ONE;
     }
--- a/src/pot.sol
+++ b/src/pot.sol
@@ -129,6 +129,7 @@ contract Pot is DSNote {
     }
 
     function cage() external note auth {
+        require(live == 1);
         live = 0;
         dsr = ONE;
     }
--- a/src/end.sol
+++ b/src/end.sol
@@ -57,7 +57,6 @@ contract PotLike {
     function cage() external;
 }
 contract VowLike {
-    function heal(uint256 rad) external;
     function cage() external;
 }
 contract Flippy {
@@ -86,6 +85,7 @@ contract Spotty {
     }
     function par() external view returns (uint256);
     function ilks(bytes32) external view returns (Ilk memory);
+    function cage() external;
 }
 
 /*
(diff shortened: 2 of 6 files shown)


src/pot.sol
```diff
@@ -129,6 +129,7 @@ contract Pot is DSNote {
     }
 
     function cage() external note auth {
+        require(live == 1);
```
gbalabasquer: why do we need this?

gbalabasquer:
Why is there a need to have `spot` with a `cage` method? Trying to understand which case we want to avoid by adding this.

src/pot.sol
gbalabasquer: The only one that seems necessary to prevent a double running of `cage` is in the `vow` (which is added), but here is not necessary.

gbalabasquer:
> Why is there a need to have `spot` with a `cage` method? Trying to understand which case we want to avoid by adding this.

Nvm, it uses the `spot.par` value.

src/pot.sol
kmbarry1: Fair enough, this one really isn't necessary, I guess I was just feeling extra paranoid when I worked on this file :). Will remove it.

kmbarry1: done

gbalabasquer:
Regarding not adding the check in `vow.rely`. I'm wondering if there is a way to mess something up in `end` adding value to `vow.Sin`. I couldn't find anything until now, but did you explore that path?

kmbarry1:
Yes, I did think about that, and at one point even had a check in `Vow.rely`. I didn't see any clear issue with it. But thinking more actually it might be used to block `heal` (which isn't directly called by End but might need to be called by a settlement keeper) maybe in conjunction with a long `wait` to block `flog`.

I'll add it back, but just in `rely` (still don't see any harm with allowing `deny` during ES).

kmbarry1:
(arguably maybe we should block `fess` during shutdown instead?)

gbalabasquer:
Hmmm regardless it is true, that a previously relied contract could access to it, I think that would be part of a malicious governance attack, which should have previously made people to run emergency shutdown.
I'd prefer to block `rely` as I think it is more in line what this PR is about and as you said we'd avoid that extra cost of gas in each `fess` operation. But yeah it is a personal taste.

kmbarry1:
Okay, I added it. This is, after all, a hardening PR, not a bulletproofing PR, and liquidations are not a "rare" event the way updating governance parameters are.

gbalabasquer (approved):
LGTM. You might want to squash the commits on merging.

PR #84, opened 2019-10-25
deauthorize obsolete collateral auction contracts

kmbarry1:
The Cat must authorize Flip contracts via `hope` to move gem from the Cat to the Flip when an auction is kicked. When a Flip is replaced, the old one should be deauthorized from manipulating the Cat's balances in the Vat as a security precaution.


kmbarry1:
Although...it occurs to me only after creating this PR that we could just move the `flux` call into `Cat.bite` which eliminates the need for authorization entirely.

gbalabasquer:
Yes you are right, moving `flux` to `cat.bite` seems to simplify it. And I like the idea to not have to set up the `hope` thing.

gbalabasquer:
Do you want to create a clean PR?

kmbarry1:
Yeah, that's a little neater. Will send one shortly.

PR #85, opened 2019-10-25
Send collateral to Flipper in bite()

kmbarry1:
This removes the need to authorize the Flipper to manipulate the Cat's Vat balances via `hope` in Cat's file method for changing the Flipper of a given ilk.


gbalabasquer (approved):
The PR looks good but reviewing it actually implemented and with the unit tests changes, feels a bit strange the `flip` kicks the auction without making sure the collateral is there.
Anyway I'm approving the PR as the technical part is correct and it is actually simpler but let's see if someone else has something to say about it.

PR #88, opened 2019-11-01
Use Lib- prefix for lib.sol:DSNote to prevent conflicts

brianmcmichael:
The `lib.sol` includes a `DSNote` contract that conflicts with the `dapptools/DSNote` contract.

This becomes an issue when flattening out the `dss-deploy` repo, which depends on the dapptools DSNote contract for `DSPause` and `ESM`, and this `lib.sol:DSNote` for everything else.

We could also add a symbolic prefix in the flattener at build time to prevent the conflict, but that would create a different bytecode output that wouldn't be subject to our FV processes.

Ideally we'd want to have our contract class names match up with our `sol` file names, but since this looks to be a library class I think the simplest solution would be to just add the prefix here.

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -46,7 +46,7 @@ contract VowLike {
     function fess(uint) external;
 }
 
-contract Cat is LibDSNote {
+contract Cat is LibNote {
     // --- Auth ---
     mapping (address => uint) public wards;
     function rely(address usr) external note auth { wards[usr] = 1; }
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -46,7 +46,7 @@ contract VowLike {
     function fess(uint) external;
 }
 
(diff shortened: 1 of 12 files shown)


gbalabasquer:
I don't think it brings any issue for any dependency more than a few changes in other smart contracts.
Maybe I would just rename them to `LibNote`, but just as a personal taste.
Let's see if someone else could identify any problem.

brianmcmichael:
I like that too. Changed to `LibNote`

godsflaw (approved):

godsflaw:
This is the right solution.  The only thing it's lacking is some cool name like `MPNote` for `Maker Protocol Note`.

LGTM

gbalabasquer:
I'm merging this, I can not find anything that could generate a problem due this.

PR #91, opened 2019-11-06
Change pragmas to 0.5.12 + changes in dai.t.sol due hevm upgrade

gbalabasquer:
The changes in `dai.t.sol` have to do with a new hevm version that changes the testing account, meaning the DAI token address also changes.


iamchrissmith (approved):

kmbarry1 (commented):
Compiler upgrade this late is kinda scary, is there some feature we need from 0.5.12?

gbalabasquer:
@brianmcmichael had mentioned something yesterday and I think it was the decision to do it anyway.
If we are running the CI again then we should be safe. In the worst case, we replace the pragmas again.

kmbarry1:
Well, carry on then, but it would be nice to have clear justification for changes we're making at this point.

PR #92, opened 2019-11-06
Prevent DAI from being sent to the DAI ERC20 contract

godsflaw:
This commit will prevent `DAI` from being sent to the `DAI` `ERC20` contract.

**The problem:**

The UX for sending an ethereum transaction and an `ERC20` transaction can sometimes confuse users and cause them to send their `ERC20` tokens to the `ERC20` contract itself.

For example, `Alice` is used to sending `ETH` to `Bob`.  She frequently looks at her transaction history to get `Bob`'s address to send him more `ETH`.  Last month, she used MetaMask (or any UI) to send `DAI` to her mother `Carol` back home.  Unfortunately, this month, `Alice`'s mom `Carol` needs more money.  Being a good daughter `Alice` decides to send her mom more `DAI`; however, she's forgotten her mother's ethereum address.  No problem, she does this all the time with `Bob`.  She goes into her transaction history and pulls the recipient address from the transaction to her mother from last month.  She puts this address in as the recipient and sends her mother some `DAI`, but her mother never gets it.  What `Alice` failed to realize is that the recipient address is the `ERC20` contract for `DAI` and not her mother, `Carol`'s, address.

To those familiar with how contracts work, this is obvious.  The transaction recipient was simply the `ERC20` contract that supported the `transfer()` call.  `Carol`'s address was simply an argument to that function.

As you can see, this is a common problem with all `ERC20` tokens:
**SAI** (~**$33,057** locked): https://etherscan.io/token/0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359?a=0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359

**USDT** (~**$55,813** locked): https://etherscan.io/token/0xdac17f958d2ee523a2206206994597c13d831ec7?a=0xdac17f958d2ee523a2206206994597c13d831ec7

**BNB** (~**$23,458** locked): https://etherscan.io/token/0xB8c77482e45F1F44dE1745F52C74426C631bDD52?a=0xB8c77482e45F1F44dE1745F52C74426C631bDD52

**USDC** (~**$101,743** locked): https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48?a=0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48

**and the list goes on**: https://etherscan.io/tokens

We obviously cannot guard against all user error, but in this case the fix could not be easier, and the timing is perfect.

One last consideration.  Keep in mind that `DAI` is a collateralized token.  Lost `DAI` represents a lost claim on that collateral.  While this may be a desirable side effect of a deflationary token, it represents an extremely mild systemic risk, but perhaps is enough of a consideration for all of us to pay the gas to prevent it? 

**The solution:**

Because we're deploying a new `ERC20` for `DAI`, we have the opportunity to fix this problem.  We simply don't allow the destination address of a transfer to be the `ERC20` contract address of `DAI`.  This change will revert the transaction whenever a user makes this mistake.

**Discussion:**
* This guards some users of `DAI` against shooting themselves in the foot, by putting the gas costs onto all users of `DAI`.  We are essentially taxing the entire population of `DAI` users to protect a few.  See the "*One last consideration*" remark above.
* Can anyone think of a case where this additional gas cost would break a consumer of the interface?
* Is there a more gas efficient way to do this?
* Does someone want to do a cost analysis of how much gas this would cost in every transaction call vs. the locked balance of `DAI`.

**Test Pass:**
```
Running 35 tests for src/test/dai.t.sol:DaiTest
[PASS] testAllowanceStartsAtZero() (gas: 6252)
[PASS] testApproveSetsAllowance() (gas: 59688)
[PASS] testApproveWillModifyAllowance() (gas: 179846)
[PASS] testApproveWillNotModifyAllowance() (gas: 161590)
[PASS] testBurn() (gas: 32407)
[PASS] testBurnAuth() (gas: 158762)
[PASS] testBurnGuyAuth() (gas: 181925)
[PASS] testBurnGuyWithTrust() (gas: 171094)
[PASS] testBurnself() (gas: 33157)
[PASS] testChargesAmountApproved() (gas: 165482)
[PASS] testDaiAddress() (gas: 1269)
[PASS] testDomain_Separator() (gas: 3753)
[PASS] testFailChargeMoreThanApproved() (gas: 140025)
[PASS] testFailInsufficientFundsTransfers() (gas: 63964)
[PASS] testFailMintGuyNoAuth() (gas: 9089)
[PASS] testFailPermitAddress0() (gas: 11859)
[PASS] testFailPermitWithExpiry() (gas: 15096)
[PASS] testFailReplay() (gas: 124000)
[PASS] testFailTransferFromSelfNonArbitrarySize() (gas: 8602)
[PASS] testFailTransferToDaiContract() (gas: 68021)
[PASS] testFailTransferWithoutApproval() (gas: 66099)
[PASS] testFailUntrustedTransferFrom() (gas: 16806)
[PASS] testFailWrongAccountTransfers() (gas: 5000)
[PASS] testMintGuy() (gas: 63107)
[PASS] testMintGuyAuth() (gas: 144182)
[PASS] testMintself() (gas: 32897)
[PASS] testPermit() (gas: 114307)
[PASS] testPermitWithExpiry() (gas: 106907)
[PASS] testSetupPrecondition() (gas: 4529)
[PASS] testTransferCost() (gas: 61387)
[PASS] testTransferFromSelf() (gas: 64322)
[PASS] testTrustedTransferFrom() (gas: 145862)
[PASS] testTrusting() (gas: 77026)
[PASS] testTypehash() (gas: 3499)
[PASS] testValidTransfers() (gas: 70622)
```


brianmcmichael:
While I am not generally advocating that we protect against every potential user error, I'm tentatively in support of this one because it it so easy to do. I've attached a screen as an illustration.

Screenshot from 2019-11-06 13-41-33

Here we can see that when look at a transaction on etherscan, the tokens appear to be sent "To:" an address, which in this case is the token address. The actual token recipient address is included in smaller print and only partially. Users not familiar with how transactions operate are more likely (and do) often copy this "To" address when they revisit an old transaction to re-send tokens to a prior recipient.

We are socializing the cost for this by adding to the gas expenditure for each transaction, so I guess we need to know how much friction this adds to each transaction and decide whether it's worth it to pack this insurance into every defi transfer.

godsflaw:
This only costs an additional `74 gas` per transfer.  I compared the test run with and without this call:

**with:** [PASS] testTransferCost() (gas: 61387)
**without:** [PASS] testTransferCost() (gas: 61313)

```
61387 - 61313 = 74
```

On the entire `ERC20` contract there are `2,641,754` total transactions to date.  If we assumed all calls were some kind of transfer (they are not), then it would have cost users a collective: `195489796 gas`.  Assuming `10 gwei` that is `~1.95 ETH` spent on gas for this one check (again, assuming all transactions are transfers).  Even at peek `ETH` price of `~$1400` this would have been `~$2,730.00`.  Still nothing compared to the `~$33,057` worth of `DAI` lost.

The above is the worst case cost estimate we can come up with.

kmbarry1:
I can't help but be a little skeptical of changing our smart contract code in response to a UX bug, on a philosophical level.

brianmcmichael:
An alternate solution would be to add an extra function as an auth'ed release valve for the locked dai. That would avoid adding the cost to each transaction but would make misallocated Dai recoverable.

```
function pop(address token, uint wad) public auth note returns (bool success) {
        return ERC20(token).transfer(msg.sender, tokens);
    }
```

This code with an ERC20 interface would actually allow any ERC20 that's been accidentally sent to the contract to be pulled out by an `auth`'ed user.

nanexcool:
Not sure about `pop`. Everyone understands `mint` and `burn` now, but adding a new `auth`d function...

brianmcmichael:
I tend to agree and prefer the PR'ed solution. Just posting for discussion. The proposed code in the PR gives us some amount of protection on behalf of the dai holders while the latter could have MKR governance making auth'ed calls in response to customer service tickets, which is probably not desirable.

kmbarry1:
Dai will be lost whether we do this or not--sent to zero, lost keys, sent to other contracts, etc. It's not clear that the amount saved through this fix is significant compared to those losses. As time goes on, improved UX should reduce this particular problem.

And lastly, in MCD, governance will be able to `mint` Dai if the amount locked in the contract becomes an economic issue. Refunds could even be done periodically based on data scraped from the blockchain.

MrChico:
@kmbarry1 Governance is not authorized to the Dai contract directly, and although they can theoretically bailout users at the expense of MKR holders, doing so would clearly `suck`. 
I'm quite concerned that such proposals (including `pop` above) are being legitimised by the discussion here.

Regarding the PR at hand, I'm okay with including the extra check, but don't have strong feelings one way or another.

kmbarry1:
@MrChico I was mainly responding to the concern that if enough Dai gets locked, it could destabilize the peg. If such a thing becomes a real problem, I'd expect governance to consider all solutions, and was pointing out that there's already an obvious release valve if needed. It's a statement of technical possibility; if MKR holders don't like it, they won't do it.

godsflaw:
We will not be including this change.  While the very informal twitter poll (https://twitter.com/nanexcool/status/1192506273700425728) run by @nanexcool indicated a slight majority of support, most of the engineers involved with reviewing this either didn't feel strongly enough, or were a clear no.  Add to this the fact that this change is a late addition to `dss` and has not been reviewed by security auditors, run through the bug bounty, or formally verified and we are all leaning to close this PR.

If you're building wallet UIs or working in UX in this space, we hope this PR clearly illustrates the problem, but also indicates that there is a lack of clear support for fixing UX problems at the smart contract layer.  Almost everyone I spoke with was in full support of conditions like this being solved at the wallet or UI layer.

MrChico:
Thanks for the clarification @kmbarry1, but I fail to see how gov cant do arbitrary mints without calls to suck. Can you outline how that would happen?

kmbarry1:
@MrChico it actually depends on how the authorizations are set up.

I was imagining the governance Pause having permissions against the Dai contract, so a spell could `mint` ERC20 Dai directly w/o changing the `debt` in the Vat. Obviously this would only be done for Dai that was provably lost (e.g. sent to the Dai contract). So no change in any internal accounting balances.

If the only thing the Dai contract `relies` on is the Dai adapter (DaiJoin), what I suggested isn't possible. But I kind of expect that at least initially governance will retain `auth` access to the Dai contract, e.g. in case it is necessary to upgrade the DaiJoin contract.

MrChico:
Thankfully gov doesnt have access to the dai contract. Upgrades can be performed by swapping out the entire system, much like the scd mcd shift. Gov access to mint would be quite detrimental

MrChico:
More specifically it would violate the fundamental equation of dai

kmbarry1:
What do you mean by "fundamental equation of dai"? `Dai.totalSupply() == Vat.debt()`?

MrChico:
See https://hackmd.io/lWCjLs9NSiORaEzaWRJdsQ and https://github.com/runtimeverification/mkr-mcd-spec

PR #93, opened 2019-11-06
Revert messages

gbalabasquer:
Please check the wording, you might do a much better job than me there.

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -99,16 +99,16 @@ contract Cat is LibNote {
     // --- Administration ---
     function file(bytes32 what, address data) external note auth {
         if (what == "vow") vow = VowLike(data);
-        else revert("Cat/file-wrong-param");
+        else revert("Cat/file-unrecognized-param");
     }
     function file(bytes32 ilk, bytes32 what, uint data) external note auth {
         if (what == "chop") ilks[ilk].chop = data;
         else if (what == "lump") ilks[ilk].lump = data;
-        else revert("Cat/file-wrong-param");
+        else revert("Cat/file-unrecognized-param");
     }
     function file(bytes32 ilk, bytes32 what, address flip) external note auth {
         if (what == "flip") { ilks[ilk].flip = flip; vat.hope(flip); }
-        else revert("Cat/file-wrong-param");
+        else revert("Cat/file-unrecognized-param");
     }
 
     // --- CDP Liquidation ---
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -47,7 +47,10 @@ contract Cat is LibNote {
     mapping (address => uint) public wards;
     function rely(address usr) external note auth { wards[usr] = 1; }
     function deny(address usr) external note auth { wards[usr] = 0; }
-    modifier auth { require(wards[msg.sender] == 1); _; }
+    modifier auth {
+        require(wards[msg.sender] == 1, "Cat/not-authorized");
+        _;
+    }
 
     // --- Data ---
     struct Ilk {
@@ -96,16 +99,16 @@ contract Cat is LibNote {
     // --- Administration ---
     function file(bytes32 what, address data) external note auth {
         if (what == "vow") vow = VowLike(data);
-        else revert();
+        else revert("Cat/file-wrong-param");
     }
     function file(bytes32 ilk, bytes32 what, uint data) external note auth {
         if (what == "chop") ilks[ilk].chop = data;
         else if (what == "lump") ilks[ilk].lump = data;
-        else revert();
+        else revert("Cat/file-wrong-param");
     }
     function file(bytes32 ilk, bytes32 what, address flip) external note auth {
         if (what == "flip") { ilks[ilk].flip = flip; vat.hope(flip); }
-        else revert();
+        else revert("Cat/file-wrong-param");
     }
 
     // --- CDP Liquidation ---
@@ -113,13 +116,13 @@ contract Cat is LibNote {
         (, uint rate, uint spot) = vat.ilks(ilk);
         (uint ink, uint art) = vat.urns(ilk, urn);
 
-        require(live == 1);
-        require(spot > 0 && mul(ink, spot) < mul(art, rate));
+        require(live == 1, "Cat/not-live");
+        require(spot > 0 && mul(ink, spot) < mul(art, rate), "Cat/not-unsafe");
 
         uint lot = min(ink, ilks[ilk].lump);
         art      = min(art, mul(lot, art) / ink);
 
-        require(lot <= 2**255 && art <= 2**255);
+        require(lot <= 2**255 && art <= 2**255, "Cat/overflow");
         vat.grab(ilk, urn, address(this), address(vow), -int(lot), -int(art));
 
         vow.fess(mul(art, rate));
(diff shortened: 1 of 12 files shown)


src/cat.sol
```diff
@@ -96,30 +99,30 @@ contract Cat is LibNote {
     // --- Administration ---
     function file(bytes32 what, address data) external note auth {
         if (what == "vow") vow = VowLike(data);
-        else revert();
+        else revert("Cat/file-wrong-param");
```
kmbarry1: nit: "unrecognized" instead of "wrong" is a little more descriptive, but not a big deal if you're trying to prioritize brevity

src/cat.sol
```diff
@@ -96,30 +99,30 @@ contract Cat is LibNote {
...
 
     // --- CDP Liquidation ---
     function bite(bytes32 ilk, address urn) external returns (uint id) {
         (, uint rate, uint spot) = vat.ilks(ilk);
         (uint ink, uint art) = vat.urns(ilk, urn);
 
-        require(live == 1);
-        require(spot > 0 && mul(ink, spot) < mul(art, rate));
+        require(live == 1, "Cat/not-live");
+        require(spot > 0 && mul(ink, spot) < mul(art, rate), "Cat/not-unsafe");
```
kmbarry1: Wouldn't "not-safe" be more correct?

src/end.sol
```diff
@@ -293,56 +296,56 @@ contract End is LibNote {
 
         uint art = tab / rate;
         Art[ilk] = add(Art[ilk], art);
-        require(int(lot) >= 0 && int(art) >= 0);
+        require(int(lot) >= 0 && int(art) >= 0, "End/ink-art-zero");
```
kmbarry1: "End/ink-art-int-overflow" since technically this is checking whether `lot` and `art` fit in an `int256`?

src/end.sol
```diff
@@ -293,56 +296,56 @@ contract End is LibNote {
...
     }
 
     function thaw() external note {
-        require(live == 0);
-        require(debt == 0);
-        require(vat.dai(address(vow)) == 0);
-        require(now >= add(when, wait));
+        require(live == 0, "End/still-live");
+        require(debt == 0, "End/debt-not-zero");
+        require(vat.dai(address(vow)) == 0, "End/dai-not-zero");
```
kmbarry1: "End/surplus-not-zero"?

src/flap.sol
```diff
@@ -112,19 +115,19 @@ contract Flapper is LibNote {
         emit Kick(id, lot, bid);
     }
     function tick(uint id) external note {
-        require(bids[id].end < now);
-        require(bids[id].tic == 0);
+        require(bids[id].end < now, "Flapper/not-finished");
+        require(bids[id].tic == 0, "Flapper/already-bidded");
```
kmbarry1: "Flapper/bid-already-placed"?

src/flap.sol
```diff
@@ -112,19 +115,19 @@ contract Flapper is LibNote {
...
         bids[id].end = add(uint48(now), tau);
     }
     function tend(uint id, uint lot, uint bid) external note {
-        require(live == 1);
-        require(bids[id].guy != address(0));
-        require(bids[id].tic > now || bids[id].tic == 0);
-        require(bids[id].end > now);
+        require(live == 1, "Flapper/not-live");
+        require(bids[id].guy != address(0), "Flapper/not-set");
```
kmbarry1: "Flapper/guy-not-set"?

src/flop.sol
```diff
@@ -113,20 +116,20 @@ contract Flopper is LibNote {
         emit Kick(id, lot, bid, gal);
     }
     function tick(uint id) external note {
-        require(bids[id].end < now);
-        require(bids[id].tic == 0);
+        require(bids[id].end < now, "Flopper/not-finished");
+        require(bids[id].tic == 0, "Flopper/already-bidded");
```
kmbarry1: "Flopper/bid-already-placed"?

src/flop.sol
```diff
@@ -113,20 +116,20 @@ contract Flopper is LibNote {
...
+        require(bids[id].guy != address(0), "Flopper/not-set");
+        require(bids[id].tic > now || bids[id].tic == 0, "Flopper/already-finished-tic");
+        require(bids[id].end > now, "Flopper/already-finished-end");
 
-        require(bid == bids[id].bid);
-        require(lot <  bids[id].lot);
-        require(mul(beg, lot) <= mul(bids[id].lot, ONE));
+        require(bid == bids[id].bid, "Flopper/not-matching-bid");
+        require(lot <  bids[id].lot, "Flopper/lot-not-lower");
+        require(mul(beg, lot) <= mul(bids[id].lot, ONE), "Flopper/not-minimum-decrease");
```
kmbarry1: "Flopper/insufficient-decrease"?
or
"Flopper/decrease-too-small"?

src/vat.sol
```diff
@@ -177,19 +180,19 @@ contract Vat {
         debt     = add(debt, dtab);
 
         // either debt has decreased, or debt ceilings are not exceeded
-        require(either(dart <= 0, both(mul(ilk.Art, ilk.rate) <= ilk.line, debt <= Line)));
+        require(either(dart <= 0, both(mul(ilk.Art, ilk.rate) <= ilk.line, debt <= Line)), "Vat/ceiling-exceed");
```
kmbarry1: "ceiling-exceeded" or similar per Mariano's suggestion

src/vow.sol
```diff
@@ -101,40 +104,40 @@ contract Vow is LibNote {
...
+        require(rad <= vat.dai(address(this)), "Vow/not-enough-dai");
         Ash = sub(Ash, rad);
         vat.heal(rad);
     }
 
     // Debt auction
     function flop() external note returns (uint id) {
-        require(sump <= sub(sub(vat.sin(address(this)), Sin), Ash));
-        require(vat.dai(address(this)) == 0);
+        require(sump <= sub(sub(vat.sin(address(this)), Sin), Ash), "Vow/debt-lower-sump");
```
kmbarry1: "Vow/insufficient-debt"?

kmbarry1: Or "Vow/not-enough-debt", to be consistent with "not-enough-surplus" below (actually I'd go w/"insufficient" for both, but it's not too big of a deal).

kmbarry1 (approved):
Overall LGTM, but made some optional suggestions to make a few of the messages a little clearer/more natural.

src/cat.sol
gbalabasquer: Good, I'll change it.

src/cat.sol
gbalabasquer: The condition is that the Vault needs to be unsafe to be bitten.

src/end.sol
gbalabasquer: Oh yeah you are right, this should be an "overflow" message

src/cat.sol
iamchrissmith: except it is checking that it is "unsafe" in order to bite it.  maybe `Cat/safe-no-bite` or `Cat/must-be-unsafe`?

src/flap.sol
iamchrissmith: This is actually checking that the auction is live/initiated, i.e. that you're not bidding on a `0` auction. So maybe
`require(live == 1, "Flapper/flapper-not-live");`
and 
`require(bids[id].guy != address(0), "Flapper/auction-not-live");`

PR #94, opened 2019-11-06
restructuring mul() call

godsflaw:
This flips args of `mul()` for some small gas savings.  We know from line `169` that `ilk.rate` can never be `0`; however, there are many potential times `urn.art` will be `0`.  In the implementation of `mul()` on line `114` we short-circuit the `require()` if `y == 0` thus saving some gas.

**bonus**: The readability is slightly better because of how the arguments are ordered in the call above.

This may be too small an optimization too late in the cycle, but we thought we would bring it up.  Feel free to close.

### DIFF
--- a/src/vat.sol
+++ b/src/vat.sol
@@ -173,7 +173,7 @@ contract Vat {
         ilk.Art = add(ilk.Art, dart);
 
         int dtab = mul(ilk.rate, dart);
-        uint tab = mul(urn.art, ilk.rate);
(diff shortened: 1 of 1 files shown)


gbalabasquer:
I don't see any harm in merging this now. @kmbarry1 @nanexcool ?

nanexcool:
> I don't see any harm in merging this now. @kmbarry1 @nanexcool ?

I like it

Issue #89, opened 2019-11-02
Return chi from pot.drip()


epheph:
Related to #87 (potentially incompatible with)

If you already have the pot's `chi` variable loaded from storage, it is worthwhile to return that value. Many (most?) external callers are likely to need that value immediately after calling `pot.drip()` (and will need to get that value in a subsequent call). Returning the data already loaded from storage would cost very little gas, a secondary call will cost ~2,000 gas

epheph:
Nice! 👍

PR #95, opened 2019-11-06
Add return value for jug.drip and pot.drip

gbalabasquer:


nanexcool (approved):
This looks good. Is it for anything in particular? Thinking mostly for web3 calls or other contracts?

gbalabasquer:
@nanexcool for other contracts, comes from this issue: https://github.com/makerdao/dss/issues/89

epheph:
Ya, if you have a contract that wants to store your DAI in DSR, you need to drip() before you join/exit the pot, and you often need to know the conversion rate, so you'll need to drip, THEN call chi (to read back out of storage).

PR #99, opened 2019-11-08
ISSUE-87: Return early if now == rho

### LINKED ISSUE #87: Return early if `now == rho`.
https://github.com/makerdao/dss/blob/effdda3657f71fd6efc3465dc661b375d1bacc3e/src/pot.sol#L138

This function is a no-op if `rho == now`.  Rather than wasting a bunch of effort (gas) executing this function in that case, it should just return early.  While the caller can do `if (pot.rho != now) pot.drip()`, it is much more expensive for them to do this than it is for this method to do that.

godsflaw:
This is fixes #87 

gas saving are `40,053` if called in the same block.

### DIFF
--- a/src/pot.sol
+++ b/src/pot.sol
@@ -137,14 +137,15 @@ contract Pot is LibNote {
     }
 
     // --- Savings Rate Accumulation ---
-    function drip() external note returns (uint tmp) {
+    function drip() external note returns (uint) {
         if (now == rho) return chi;
         require(now > rho, "Pot/invalid-now");
-        tmp = rmul(rpow(dsr, now - rho, ONE), chi);
+        uint tmp = rmul(rpow(dsr, now - rho, ONE), chi);
         uint chi_ = sub(tmp, chi);
         chi = tmp;
         rho = now;
         vat.suck(address(vow), address(this), mul(Pie, chi_));
+        return chi;
     }
 
     // --- Savings Dai Management ---
--- a/src/pot.sol
+++ b/src/pot.sol
@@ -138,7 +138,8 @@ contract Pot is LibNote {
 
     // --- Savings Rate Accumulation ---
     function drip() external note returns (uint tmp) {
-        require(now >= rho, "Pot/invalid-now");
+        if (now == rho) return chi;
+        require(now > rho, "Pot/invalid-now");
         tmp = rmul(rpow(dsr, now - rho, ONE), chi);
         uint chi_ = sub(tmp, chi);
         chi = tmp;
--- a/src/jug.sol
+++ b/src/jug.sol
@@ -98,11 +98,13 @@ contract Jug is LibNote {
     }
 
     // --- Stability Fee Collection ---
-    function drip(bytes32 ilk) external note returns (uint rate) {
+    function drip(bytes32 ilk) external note returns (uint) {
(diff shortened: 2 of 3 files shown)


src/pot.sol
```diff
@@ -138,7 +138,8 @@ contract Pot is LibNote {
 
     // --- Savings Rate Accumulation ---
     function drip() external note returns (uint tmp) {
-        require(now >= rho, "Pot/invalid-now");
+        if (now == rho) return chi;
+        require(now > rho, "Pot/invalid-now");
```
MicahZoltu: Unrelated to this PR, shouldn't this be `invalid-rho`?

godsflaw: time is relative...  Sorry, been a long day.

kmbarry1:
Just to accurately represent the calculus behind this, using the following numbers (which others measured, I have not checked them):

savings of optimization when `rho == now`: 40,053 gas
cost of optimization when `rho != now`: 300 gas

one finds that the odds of calling `drip` when `rho == now` need to be about 1/134 for this "optimization" to actually be an improvement, on average.

Making this change represents a certain level of faith that the system will reach a scale where this condition holds so real gas savings accrue. Although my priors aren't terribly well-informed, I suspect that this will be a net loss at least for some period of time (I'm expecting << 1 `drip` call per block, at least initially, e.g. on the order of a few per hour).

Of course, EVM storage operations tend to get more expensive with each hard fork, so it is more likely that the threshold become more favorable over time than not.

brianmcmichael:
It's sufficiently complex and the rewards so minimal as to be unlikely, but as MCD scales this mechanism could theoretically be utilized by miners to bundle these transactions (or unbundle them) to consume less (or capture additional) gas for each call. I don't think this side-effect is an argument for or against, but thought it was worth pointing out the observation.

MicahZoltu:
I think something that isn't being considered in this discussion is that if this change is _not_ mode, then anyone who calls `drip` _MUST_ do `if (pot.rho() != now) pot.drip()`.  If they don't, they risk a failed transaction.  It is significantly more expensive for an external caller to do that check than it is to do it internally, especially since internally you are already loading `rho` from storage.

Given that all callers must first do the check, I believe this change is a gas reduction in _all_ cases except the theoretical (but likely never encountered) case where the caller knows they are at the top of the block or otherwise has some guarantee that no transactions before it have triggered a `pot.drip`.

gbalabasquer:
why do they risk a failed transaction? If equals the tx still passes without this change, just consumes more gas.

MicahZoltu:
Ah, sorry.  Was thinking of something else.  I shouldn't comment on GitHub late at night.

MicahZoltu:
It is possible that if the DSR is high enough, bots will `drip` regularly as they move capital in/out of pot so they get the savings rate, but they need DAI to do their bot stuff.  This is a potentiality where this function is hit with some regularity, enough to make the gas savings worth it.

godsflaw:
Sorry to say, while we think the gains here are worthy, we are just taking too much risk to include this now.  We have done multiple audits, a bug bounty, and formal verification of the system and after much deliberation we believe even a small change like this can invalidate those results and threaten our confidence in the system.  Since the `pot` and `jug` are peripherally connected we can add this again later, perhaps with other optimizations, and let governance vote on upgrading these contracts after audits and formal verification.

PR #107, opened 2020-03-12
nonzero price floor Flipper

kmbarry1:
DO NOT MERGE (YET)

Only for discussion at this point. Designed to prevent extremely low bids from winning flip auctions until a more robust auction mechanism can be designed.

### DIFF
--- a/src/flip.sol
+++ b/src/flip.sol
@@ -126,7 +126,7 @@ contract Flipper is LibNote {
     }
     function file(bytes32 what, address data) external note auth {
         if (what == "spot") spot = SpotLike(data);
-        if (what == "feed") feed = FeedLike(data);
+        else if (what == "feed") feed = FeedLike(data);
         else revert("Flipper/file-unrecognized-param");
     }
 
--- a/src/flip.sol
+++ b/src/flip.sol
@@ -81,7 +81,7 @@ contract Flipper is LibNote {
 
     SpotLike public spot;  // Spotter address
     FeedLike public feed;  // This is the Medianizer, or wherever the OSM reads undelayed prices from
-    uint256  public cut;   // ratio of starting bid to market price of the lot [wad]
+    uint256  public cut;   // ratio of starting bid to market price of the lot [ray]
 
     // --- Events ---
     event Kick(
--- a/src/flip.sol
+++ b/src/flip.sol
@@ -171,9 +171,10 @@ contract Flipper is LibNote {
         if (bids[id].bid == 0) {
             // Need to be whitelisted to make this call
             (bytes32 val, bool has) = feed.peek();
-            require(has, "Flipper/no-price");
-            uint256 par = spot.par();
-            require(bid >= rmul(wmul(rdiv(uint256(val), par), lot), cut));
+            if (has) {
+                uint256 par = spot.par();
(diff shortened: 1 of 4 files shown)


src/flip.sol
```diff
@@ -111,9 +143,17 @@ contract Flipper is LibNote {
         require(kicks < uint(-1), "Flipper/overflow");
         id = ++kicks;
 
-        bids[id].bid = bid;
+        (PipLike pip,) = spot.ilks(ilk);
+
+        // Need to be whitelisted to make this call
```
kmbarry1: If we don't want to whitelist the Flipper to cal the OSM, we could instead do math based on `spot` in the Vat and the `mat` value.

andy8052 (approved):

src/flip.sol
```diff
@@ -138,6 +167,16 @@ contract Flipper is LibNote {
         require(bid >  bids[id].bid, "Flipper/bid-not-higher");
         require(mul(bid, ONE) >= mul(beg, bids[id].bid) || bid == bids[id].tab, "Flipper/insufficient-increase");
 
+        // check for first bid only
+        if (bids[id].bid == 0) {
+            // Need to be whitelisted to make this call
+            (bytes32 val, bool has) = feed.peek();
+            if (has) {
+                uint256 par = spot.par();
+                require(bid >= rmul(wmul(rdiv(uint256(val), par), lot), cut));
```
gbalabasquer: So I think there is an error here (haven't fully tested it) and also want to mention one consideration:

Possible error: `val` is `WAD`, `par` is `RAY`, `lot` is `WAD`, `cut` is `WAD`. `bid` should be `RAD`.

`rmul(wmul(rdiv(WAD, RAY), WAD), WAD)`

`rdiv(WAD, RAY)` is `WAD`, so `rmul(wmul(WAD, WAD), WAD)`

`wmul(WAD, WAD)` is `WAD`, so `rmul(WAD, WAD)` => 10 ^ 9 unit not 10 ^ 45 (wad).

About the consideration, I understand `cut` as the amount you would expect to `cut` to the feed price for the first `bid`. However it acts as the opposite here. I would have expected to see `sub(ONE, cut)`. Bit less efficient though.

gbalabasquer:
In general this is how I pictured this `flip` workaround. My only concern is how this would look in `dss-deployment-scripts` as we would need to add the median to the config file as well or assume that the PIP is always an OSM and we can extract the median from it, which keeps complicating the logic about feeds there.
However if we consider it as a very temporary solution we might even keep it out from the scripts.

src/flip.sol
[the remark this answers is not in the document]
kmbarry1: Added.

kmbarry1: I actually intended `cut` to be a `ray`, the comment is wrong.
But you are correct `bid` needs to be a `rad`, not a `wad`...easy fix is to replace `rmul` with `mul`, thanks for catching!

As for naming...`cut` is far from our worst meme IMO. I can interpret it in a similar sense to its usage as a noun in phrases like "taking a cut of the profits" (here the "cut" is the amount you're getting), I think that actually matches the usage here.

gbalabasquer: Yeah ok, makes sense!

kmbarry1:
If we're willing to accept a 1 hr lag we could do it entirely with just the Spotter and Vat.

Issue #108, opened 2020-04-05
Debt Auctions Can Be Blocked


kmbarry1:
# Fixing Debt Auction Blocking

## Background

If surplus collected from a debt auction is canceled with free unbacked debt (`woe`) using `Vow.heal` instead of with on-auction debt (`Ash`) via `kiss`, future debt auctions (`flops`) can be prevented from occurring (see [these slides](https://docs.google.com/presentation/d/1nnpPBiOLEWi81q8zrHoIWH4s3iQaKrCSaR68AafzQJo/edit#slide=id.p) for an in-depth explanation; discussion occurred in [this governance call](https://youtu.be/erh25lnaIo0?t=376)). While governance can take action to unblock these debt auctions, it would be much better if the system either prevented this situation entirely or allowed it to be resolved without a governance vote.

## Primary Fix Options

While far from exhaustive, the options presented here cover what are probably the simplest and least disruptive fix options, and are thus the main ones under debate.

### Instant Access Module

When an amount `zed` of debt becomes "stuck" as `Ash`, governance can unblock the stuck debt by converting it into `woe` via calling first `Vat.suck(address(Vow), address(Vow), zed)` and then `Vow.kiss(zed)`. An appropriately permissioned (`relied` upon by the Vat) instant access module could enable this operation to be be done by anyone at any time, without requiring a governance vote.

The obvious attractive feature of this solution is that it requires no replacements of any existing core contracts. It is complicated, however, by the need to avoid "unsticking" auction debt that has not yet received a bid (otherwise extra, unnecessary debt auctions could occur). The simplest solution to this issue is to have the module keep track of how much debt is up for auction but as-of-yet unbid upon. This can be done without loops in the contract itself, for example as follows:

    // Define here mock interfaces for Vat, Vow, and Flopper.
    
    contract FlopUnblocker {
        // Define here obviously needed things like a constructor,
        // safe math functions, etc.
    
        Vat public vat;
        Vow public vow;
        Flopper public flop;
    
        mapping (uint => uint) public known;
        uint public knownAuctionCount;
        mapping (uint => uint) public ashes;
        uint public trueAsh;
    
        function unblock() public {
            require(knownAuctionCount == flop.kicks());
    
            // kiss away as much debt as possible
            uint joy = vat.dai(address(vow));
            require(joy < vow.Ash());  // optional: could instead kiss away Ash
            vow.kiss(joy);
    
            // unblock stuck auctions
[...]

PR #110, opened 2020-04-13
kiss gal in first dent call

kmbarry1:
Live-system fix for debt auction blocking: https://github.com/makerdao/dss/issues/108

### DIFF
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -103,7 +103,7 @@ contract Flopper is LibNote {
     function file(bytes32 what, uint data) external note auth {
         if (what == "beg") beg = data;
         else if (what == "pad") pad = data;
-        else if (what == "ttl") ttl = uint48(data);
+        else if (what == "ttl") { require(uint48(data) > 0); ttl = uint48(data); }
         else if (what == "tau") tau = uint48(data);
         else revert("Flopper/file-unrecognized-param");
     }
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -106,7 +106,7 @@ contract Flopper is LibNote {
     function file(bytes32 what, uint data) external note auth {
         if (what == "beg") beg = data;
         else if (what == "pad") pad = data;
-        else if (what == "ttl") { require(uint48(data) > 0); ttl = uint48(data); }
+        else if (what == "ttl") ttl = uint48(data);
         else if (what == "tau") tau = uint48(data);
         else revert("Flopper/file-unrecognized-param");
     }
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -27,6 +27,7 @@ contract GemLike {
     function mint(address,uint) external;
 }
 contract VowLike {
+    function Ash() public returns (uint);
     function kiss(uint) external;
 }
 
@@ -140,7 +141,12 @@ contract Flopper is LibNote {
 
         // on first dent, clear Ash
         if (bids[id].tic == 0) {
-            VowLike(bids[id].guy).kiss(bid);
+            uint Ash = VowLike(bids[id].guy).Ash();
+            if (Ash >= bid) {
+                VowLike(bids[id].guy).kiss(bid);
+            } else {
+                VowLike(bids[id].guy).kiss(Ash);
+            }
         }
 
         bids[id].guy = msg.sender;
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -26,6 +26,9 @@ contract VatLike {
 contract GemLike {
     function mint(address,uint) external;
 }
+contract VowLike {
+    function kiss(uint) external;
+}
 
 /*
    This thing creates gems on demand in return for dai.
@@ -134,6 +137,11 @@ contract Flopper is LibNote {
         require(mul(beg, lot) <= mul(bids[id].lot, ONE), "Flopper/insufficient-decrease");
 
         vat.move(msg.sender, bids[id].guy, bid);
+
+        // on first dent, clear Ash
+        if (bids[id].tic == 0) {
+            VowLike(bids[id].guy).kiss(bid);
+        }
 
         bids[id].guy = msg.sender;
         bids[id].lot = lot;
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -142,15 +142,16 @@ contract Flopper is LibNote {
 
         if (msg.sender != bids[id].guy) {
             vat.move(msg.sender, bids[id].guy, bid);
+
+            // on first dent, clear as much Ash as possible
+            if (bids[id].tic == 0) {
+                uint Ash = VowLike(bids[id].guy).Ash();
+                VowLike(bids[id].guy).kiss(min(bid, Ash));
+            }
+
+            bids[id].guy = msg.sender;
         }
 
-        // on first dent, clear as much Ash as possible
-        if (bids[id].tic == 0) {
-            uint Ash = VowLike(bids[id].guy).Ash();
-            VowLike(bids[id].guy).kiss(min(bid, Ash));
-        }
-
-        bids[id].guy = msg.sender;
         bids[id].lot = lot;
         bids[id].tic = add(uint48(now), ttl);
     }
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -98,6 +98,9 @@ contract Flopper is LibNote {
     function mul(uint x, uint y) internal pure returns (uint z) {
         require(y == 0 || (z = x * y) / y == x);
     }
+    function min(uint x, uint y) internal pure returns (uint z) {
+        if (x > y) { z = y; } else { z = x; }
+    }
 
     // --- Admin ---
     function file(bytes32 what, uint data) external note auth {
@@ -139,14 +142,10 @@ contract Flopper is LibNote {
 
         vat.move(msg.sender, bids[id].guy, bid);
 
-        // on first dent, clear Ash
+        // on first dent, clear as much Ash as possible
         if (bids[id].tic == 0) {
             uint Ash = VowLike(bids[id].guy).Ash();
-            if (Ash >= bid) {
-                VowLike(bids[id].guy).kiss(bid);
-            } else {
-                VowLike(bids[id].guy).kiss(Ash);
-            }
+            VowLike(bids[id].guy).kiss(min(bid, Ash));
         }
 
         bids[id].guy = msg.sender;
--- a/src/test/flop.t.sol
+++ b/src/test/flop.t.sol
@@ -137,6 +137,39 @@ contract FlopTest is DSTest {
         assertEq(vat.dai(ali), 190 ether);
         // gal receives payment
         assertEq(vat.dai(gal),  10 ether);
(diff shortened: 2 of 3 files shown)


src/flop.sol
```diff
@@ -135,6 +138,11 @@ contract Flopper is LibNote {
 
         vat.move(msg.sender, bids[id].guy, bid);
 
+        // on first dent, clear Ash
+        if (bids[id].tic == 0) {
+            VowLike(bids[id].guy).kiss(bid);
```
gbalabasquer, line 143: I would actually check `Ash` as you had previously suggested. It is true that in normal operations shouldn't happen to not have `Ash` when doing the first `bid`. But wondering if a possible future `Vow` replacement would bring issues here.
If we do the full replacement with `flop` then we should be ok, but if we want to maintain the `flop` for any reason, there we would have an issue.
I'm not really sure about this, but wondering if the extra check might be beneficial just in case. It's true that will add some nasty extra gas. We will have an external call and storage load.

kmbarry1: No, you're absolutely right, it is not correct if it does not check the `Ash`. Otherwise, someone could `kiss` a little bit of `Ash` before the first `dent` and block `denting`.

gbalabasquer: Ah that's true, you can still `kiss` from generated surplus via the fees or `flip` auctions.

kmbarry1: Fixed it.

gbalabasquer (approved):
LGTM

src/flop.sol
```diff
@@ -135,6 +139,16 @@ contract Flopper is LibNote {
...
+            if (Ash >= bid) {
+                VowLike(bids[id].guy).kiss(bid);
+            } else {
+                VowLike(bids[id].guy).kiss(Ash);
+            }
+        }
+
         bids[id].guy = msg.sender;
         bids[id].lot = lot;
         bids[id].tic = add(uint48(now), ttl);
```
ehildenb, line 154: This makes modelling this code harder, because a state update happen _after_ a potential call to `Vow.kiss`. So I would need to split `dent` up into two steps, so that I can do `Flop.dent_init`, then `Vow.kiss`, then `Flop.dent_finish` to finish the state update.

Is there any chance that the added clearing of the ash can happen _after_ the state updates to this contract (the updates to `bid[id] ...`? I think it would only require storing `origGuy = bids[id].guy`. Either way, that's likely how I'll do it in the K.

Also, can the internal:

```
            if (Ash >= bid) {
                VowLike(bids[id].guy).kiss(bid);
            } else {
                VowLike(bids[id].guy).kiss(Ash);
            }
```

Be replaced with a single line:

```
VowLike(bids[id].guy).kiss(min(bid, Ash))
```

ehildenb: This also, unfortunately, creates a circular dependency between `Flop` and `Vow`.

ehildenb: I'm also looking at `VowLike(bids[id].guy)`, which looks quite weird to me. I guess we're relying on the fact that only the `Vow` can `kick` an auction, so when `tic == 0`, then we know that the `guy` is the `Vow`.

But there also are no guards against the `ttl` being set to 0, so you could conceivably have an admin set `TTL` to `0`, then a normal `Flop . dent` would go through and change `bids[id].guy` to be something other than `Vow` (whoever calls `Flop . dent`). AFAICT, this would cause the call to `kiss` to fail, and then the `dent` too fail, so then `dent`ing that auction would be blocked by the fact that the `kiss` can't go through.

ehildenb: Here is the structural change I need to make to the K to avoid the circular dependency: https://github.com/makerdao/mkr-mcd-spec/commit/eea7db11f02f07eb300e50e8903edad14eed5aad

And here is the change I make to `dent` to implement this fix: https://github.com/makerdao/mkr-mcd-spec/commit/bdd3d177acbd91c107055f7e5a8df4f5a0eb383e

ehildenb: We could also require `ttl > 0` in the `file` method.

ehildenb: Another option would be to make the `vow-addr` a `file`able parameter, and use that instead of `bids[id].guy`. This would also eliminate the problem with the `ttl == 0`, because we wouldn't be relying on the first bid being from the `Vow`.

ehildenb: If it was file-able, we could just check that the `guy == vow-addr`, instead of comparing with `tic`.

kmbarry1: Fair point about `ttl`--while hopefully governance would never make the mistake of setting it to zero (this makes auctions `undealable`, so it would break the contract even without this change), and malicious governance can do _much worse_ things (e.g. steal all collateral), it doesn't hurt to foolproof a bit and require a non-zero `ttl` in `file`.

Adding a `file` capability for the Vow's address was considered when fixing a previous issue (`yank` bug) and is worth considering. The worst thing about it is that it creates this tension where there's an overwhelming desire to simplify the code--`require(msg.sender == vow);` in `kick`, get rid of the `gal` argument to `kick`, etc. It also comes with it's own misconfiguration risk. But it's not a bad option, especially since the Vow's address is needed post-`cage` (and right now is obtained by assuming the Vow is the caller of `cage`). Will think about this more.

Re: `min`...it doesn't save many LoC since there's only one usage and makes the code less local, but OTOH I guess everyone knows what a function named `min` does so it doesn't really decrease readability.

ehildenb: Looks good to me!

ehildenb:
This looks good to me, but the build is failing, and someone with more Solidity experience should also peruse the changes.

kmbarry1:
Looks like the build error is due to some Nix/dapptools interaction. All tests pass locally.

andy8052 (approved):
Also working for me locally

andy8052:
If we are requiring `ttl > 0` do we also want to require `tau > 0`. It is just logically weird if the max auction time can be shorter than the bid duration in my mind

kmbarry1:
> If we are requiring ttl > 0 do we also want to require tau > 0. It is just logically weird if the max auction time can be shorter than the bid duration in my mind

Hmmm...I mean, at a certain point, you have to trust governance not to be totally moronic. Additionally, there's not really a practical difference between `tau == 0` and, say `tau == 1` (or `tau == <anything less than the typical time in seconds between blocks>`). The contract also doesn't rely quite as strongly on `tau =/= 0` as it does on `ttl =/= 0` (arguably, it doesn't rely on `tau =/= 0` at all). And this sort of parameter constraint logic can always be encapsulated in a `Mom` contract later on, I don't think it needs to clutter the core protocol. I made an exception for `ttl` because the contract and system as a whole are now depending quite heavily on `ttl =/= 0` for correct behavior; I just don't think `tau` meets the same standard.

gbalabasquer:
Guys can you explain me better why `ttl == 0` provokes a failure?
If `ttl == 0`, then first `dent` should pass, and should clean the `Ash` with `kiss`.

The only effect I see here is that the auction will be finished after the first `dent`.

I do not see that another address than the `vow` will call `kiss` as `now + 0` is `now` and not `0`, then the condition `tic != 0` will not pass again.

ehildenb:
That's a good point that `tic` will be set to `now + ttl`, so as long as `now` is always greater than `0` it should be fine without the constraint on `ttl`.

gbalabasquer:
Regarding the constraint, there is actually a difference between  setting `tau` or `ttl` to 0 and 1. The difference in practice is that setting them to 0 makes that only the first `dent` passes. Setting it to 1, makes that more than one `dent` can pass in the same block.
Anyway I don't consider this situation is worth to add the constraint.

kmbarry1:
Oh yeah I got confused...probably not worth having the `require` after all.

gbalabasquer:
I've rebased master which includes a fix for the CI.

kmbarry1:
Btw I am leaning towards going ahead and merging this as is. I've debated in my head a lot both 1) maybe we should replace the Vow instead (too much technical risk IMO), and 2) should we add an explicit governance parameter for the Vow to the Flopper (IMO no, there's no significant advantage to that).

So speak up soon if you feel differently :).

gbalabasquer:
I think is fine as it is. But do we want to propose the change of the flop soon?
If that is happening, we might also consider merging https://github.com/makerdao/dss/pull/109
I'm also wondering if we should wait for the executive to pass before merging the changes.

kmbarry1:
I strongly support merging and deploying #109 together.

There is no rush, I probably wasn't going to actually merge until next Monday (although I'm not too sure why we need to wait for the executive to pass).

gbalabasquer:
I guess it has more to do with discussing how we follow changes in the repos against what we have deployed on mainnet. But this is part a more broad conversation.
But eventually these two PRs should be included in a PR and passed, they aren't controversial changes.

gbalabasquer:
So I had to merge the `staging` branch which includes the double funds fix here to resolve some conflicts. Doing it I've realized we can not have the `bids[id].guy` value being updated in the same `if` conditional that checks `msg.sender != bids[id].guy` as if we place it before the `tic == 0` conditional, the `bids[id].guy` will not be `vow` anymore and if we place the whole thing after, there might not be surplus to `kiss` the `bid` (due the missing prev `vat.move`).

So had to set `bids[id].guy` back to its original position in the code.
Now that here we were forced to do it, I'm wondering if for consistency we should do it everywhere. Let me know yours thoughts. I could do it in this PR or merge it as it is and push directly another commit to `staging` changing that (better IMO). Third option is leave as it is.

kmbarry1:
Personally I do not have any real objection to leaving it how it is. I agree that if we change the other auctions to match how this logic was force to flow, that shouldn't go in this PR, but rather be its own commit to `staging`.

If we wanted to preserve the gas optimization, we could lift the `tic == 0` block into the `msg.sender != guy` block (since if `tic == 0`, `guy` is guaranteed to be changing, as the Vow does `dent`), but I think that's maybe too clever by half. I suspect successive bids by the same bidder on `flops` will much rarer than with `tend`/`dent` in `flips`.

So, up to you--I'm fine with either leaving as-is or doing a separate commit to enforce consistency.

gbalabasquer:
Well maybe having:
```
if (msg.sender != bids[id].guy) {
    vat.move(msg.sender, bids[id].guy, bid);
    // on first dent, clear as much Ash as possible
    if (bids[id].tic == 0) {
        uint Ash = VowLike(bids[id].guy).Ash();
        VowLike(bids[id].guy).kiss(min(bid, Ash));
    }
    bids[id].guy = msg.sender;
}
```
as you suggested is not actually bad and that wouldn't break my symmetry toc problem with the other auctions (lol) and would still keep the gas optimization everywhere.

kmbarry1:
Yeah, it's not terrible. I guess it really isn't strengthening assumptions about the Vow-Flopper interaction very much either (largely because those assumptions are already nearly at maximum strength). So go for it if you think it's worth it.

gbalabasquer:
Ok it's done now. I'll keep the PR open during today just in case someone wants to double check the change.
I'll get it squashed and merged to `staging` first thing tomorrow morning.

PR #111, opened 2020-04-17
Add shell.nix to make 'dapp test' not clone dapptools by itself

icetan:


kmbarry1:
@icetan is this still relevant to review+merge?

icetan:
@kmbarry1 Yes, this will also make it possible to easily acquire a local shell with the correct version of `dapp/solc` by running `nix-shell` in this repo.

shell.nix
```diff
@@ -0,0 +1,12 @@
+{ dappPkgs ? (
+    import (fetchTarball "https://github.com/makerdao/makerpkgs/tarball/master") {}
+  ).dappPkgsVersions.seth-0_8_4
```
gbalabasquer, line 3: we should be using: `hevm-0_41_0` version

icetan: For some reason `hevm-0_41_0` doesn't build due to a GitHub repo URL in the `dapptools` nix expression ending with a `/` and returning `ERROR: Repository not found.`.

This is fixed in `dapptools/master` and has been added in this PR: https://github.com/makerdao/makerpkgs/pull/16

We could bump the version of `dapptools` to include the fix.

gbalabasquer: Yeah let's ask dapphub people to bump a new version, they haven't done for several months.
When that's done, we can update this PR.
On the other hand I've seen @dizzy added to other repos (e.g. https://github.com/makerdao/dss-proxy-actions/commit/17a2b1652f15a6d9bf4b16e48ef5c4e20c4dc4a1) a docker instance to avoid compiling dapptools each time. We should standardize and do the same for all the repos. It would be nice you and George  get some consensus on the best way to run these tests, then we can apply the same thing to everyone.

kmbarry1: There's a new version now, right? Are we waiting on the consensus about how best to run tests here?

gbalabasquer:
This is actually something I forgot that existed.
So using the latest makerpkgs with the latest dapptools version already improved the execution of the tests by much. But this might something that even improves more the timing. Let me do some test, and if that's the case we will get this merged and replicated to all the other repos.

shell.nix
gbalabasquer: This is actually something I forgot that existed.
So using the latest makerpkgs with the latest dapptools version already improved the execution of the tests by much. But this might something that even improves more the timing. Let me do some test, and if that's the case we will get this merged and replicated to all the other repos.

gbalabasquer: @kmbarry1 @godsflaw @icetan @dizzy 

Comparing times for the CI in `dss`:

- `docker-build`:  3 min 8 sec (docker image)
- Actual `master`: 2 min 18 sec (nix + git cloning `dapptools`)
- This PR: 1 min 52 sec (nix)

Times are relative though, as the actual `master` had run before in 2 minutes (now it gave 18 seconds more).

So it seems the actual `master` and this PR are not very different in timing now, as the actual version of `dapptools` doesn't include submodules anymore, meaning that cloning it is quite fast (comparing to the past that needed to download submodules as well). That's why cloning `dapptools` or not is not very different now.
However it seems this PR includes the most efficient and correct way for the CI running, so personally I'd merge it and replicate this way to all the other `dapp` standard repos.

The only repos I'm not really sure how we can use this method are `spells-mainnet` and `spells-kovan` due they do testing based on a node. How that works in the docker image? as it seems to run very fast for calling an external node.

gbalabasquer:
Closing this PR as this change was done at some point in another branch.

PR #113, opened 2020-05-11
Added comments

KentonPrescott:
## Summary
This PR Improves comments within `dss` by adding units and short descriptions where needed.
Requesting reviewers check to ensure consistency and approve as soon as convenient.

## Rationale
A lack of unit awareness infects the developer experience.
Although important data structures are commented (like `Ilk` and `Urn` in the `Vat`), this is not consistent in other contracts used by integration partners, such as the `Spotter` or the `Jug`.

I hope this doesn’t spur any bike shedding, but I think it’s important to add unit context, readily available and in the source code. It saves time and will improve the integration experience immensely. 

## Changes

Following great examples in the `Vat.sol`, I tried to be as consistent with what commenting style has been used:
- Source of variable declaration within contracts - short description + units
- Imports of data structures via interface contracts - units 

For example from `Jug.sol`: 
```
contract VatLike {
    function ilks(bytes32) external returns (
        uint256 Art,   // [wad]
        uint256 rate   // [ray]
    );
    function fold(bytes32,address,int) external;
}

contract Jug is LibNote { 
… 
	struct Ilk {
	     uint256 duty;  // Stability Fee         [ray]
      	     uint256  rho;  // Time of last drip     [seconds]
    	}
… 
}
```

There were also strange variations in uppercase conditions, so I tried to improve consistency in that area as well.

When short descriptions were lacking, I looked to examples within the repo, such as other solidity files or the [Wiki glossary](https://github.com/makerdao/dss/wiki/Glossary)

I’ve also added unit context to `event Poke( ... )` in `spot.sol`, relieving a potential journey outside of `dss`.

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -26,13 +26,13 @@ contract Kicker {
 
 contract VatLike {
     function ilks(bytes32) external view returns (
-        uint256 Art,   // wad
-        uint256 rate,  // ray
-        uint256 spot   // ray
+        uint256 Art,   // [wad]
+        uint256 rate,  // [ray]
+        uint256 spot   // [ray]
     );
     function urns(bytes32,address) external view returns (
-        uint256 ink,   // wad
-        uint256 art    // wad
+        uint256 ink,   // [wad]
+        uint256 art    // [wad]
     );
     function grab(bytes32,address,address,address,int,int) external;
     function hope(address) external;
@@ -62,9 +62,9 @@ contract Cat is LibNote {
 
     mapping (bytes32 => Ilk) public ilks;
 
-    uint256 public live;
-    VatLike public vat;
-    VowLike public vow;
+    uint256 public live; // Access Flag
+    VatLike public vat;  // CDP Engine
+    VowLike public vow;  // Debt Engine
 
     // --- Events ---
     event Bite(
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -62,9 +62,9 @@ contract Cat is LibNote {
 
     mapping (bytes32 => Ilk) public ilks;
 
-    uint256 public live; // Access Flag
-    VatLike public vat;  // CDP Engine
-    VowLike public vow;  // Debt Engine
+    uint256 public live;  // Access Flag
+    VatLike public vat;   // CDP Engine
+    VowLike public vow;   // Debt Engine
 
     // --- Events ---
     event Bite(
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -62,7 +62,7 @@ contract Cat is LibNote {
 
     mapping (bytes32 => Ilk) public ilks;
 
-    uint256 public live;  // Access Flag
+    uint256 public live;  // Shutdown Flag
     VatLike public vat;   // CDP Engine
     VowLike public vow;   // Debt Engine
 
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -62,7 +62,7 @@ contract Cat is LibNote {
 
     mapping (bytes32 => Ilk) public ilks;
 
-    uint256 public live;  // Shutdown Flag
+    uint256 public live;  // Active Flag
     VatLike public vat;   // CDP Engine
     VowLike public vow;   // Debt Engine
 
--- a/src/end.sol
+++ b/src/end.sol
@@ -23,15 +23,15 @@ import "./lib.sol";
 contract VatLike {
     function dai(address) external view returns (uint256);
     function ilks(bytes32 ilk) external returns (
-        uint256 Art,
-        uint256 rate,
-        uint256 spot,
-        uint256 line,
-        uint256 dust
+        uint256 Art,   // [wad]
+        uint256 rate,  // [ray]
+        uint256 spot,  // [ray]
+        uint256 line,  // [rad]
+        uint256 dust   // [rad]
     );
     function urns(bytes32 ilk, address urn) external returns (
-        uint256 ink,
-        uint256 art
+        uint256 ink,   // [wad]
+        uint256 art    // [wad]
     );
     function debt() external returns (uint256);
     function move(address src, address dst, uint256 rad) external;
@@ -43,9 +43,9 @@ contract VatLike {
 }
 contract CatLike {
     function ilks(bytes32) external returns (
-        address flip,  // Liquidator
-        uint256 chop,  // Liquidation Penalty   [ray]
-        uint256 lump   // Liquidation Quantity  [rad]
+        address flip,
+        uint256 chop,  // [ray]
+        uint256 lump   // [rad]
     );
     function cage() external;
 }
@@ -57,14 +57,14 @@ contract VowLike {
 }
 contract Flippy {
     function bids(uint id) external view returns (
-        uint256 bid,
-        uint256 lot,
+        uint256 bid,   // [rad]
+        uint256 lot,   // [wad]
         address guy,
         uint48  tic,
         uint48  end,
         address usr,
         address gal,
-        uint256 tab
+        uint256 tab    // [rad]
     );
     function yank(uint id) external;
 }
@@ -77,7 +77,7 @@ contract Spotty {
     function par() external view returns (uint256);
     function ilks(bytes32) external view returns (
         PipLike pip,
-        uint256 mat
+        uint256 mat    // [ray]
     );
     function cage() external;
 }
@@ -193,24 +193,24 @@ contract End is LibNote {
     }
 
     // --- Data ---
-    VatLike  public vat;
+    VatLike  public vat;   // CDP Engine
     CatLike  public cat;
-    VowLike  public vow;
+    VowLike  public vow;   // Debt Engine
     PotLike  public pot;
     Spotty   public spot;
 
-    uint256  public live;  // cage flag
-    uint256  public when;  // time of cage
-    uint256  public wait;  // processing cooldown length
-    uint256  public debt;  // total outstanding dai following processing [rad]
+    uint256  public live;  // Cage flag
+    uint256  public when;  // Time of cage
+    uint256  public wait;  // Processing Cooldown Length
+    uint256  public debt;  // Total outstanding dai following processing [rad]
 
-    mapping (bytes32 => uint256) public tag;  // cage price           [ray]
-    mapping (bytes32 => uint256) public gap;  // collateral shortfall [wad]
-    mapping (bytes32 => uint256) public Art;  // total debt per ilk   [wad]
-    mapping (bytes32 => uint256) public fix;  // final cash price     [ray]
+    mapping (bytes32 => uint256) public tag;  // Cage price              [ray]
+    mapping (bytes32 => uint256) public gap;  // Collateral shortfall    [wad]
+    mapping (bytes32 => uint256) public Art;  // Total debt per ilk      [wad]
+    mapping (bytes32 => uint256) public fix;  // Final cash price        [ray]
 
-    mapping (address => uint256)                      public bag;  // [wad]
-    mapping (bytes32 => mapping (address => uint256)) public out;  // [wad]
+    mapping (address => uint256)                      public bag;  //    [wad]
+    mapping (bytes32 => mapping (address => uint256)) public out;  //    [wad]
 
     // --- Init ---
     constructor() public {
--- a/src/end.sol
+++ b/src/end.sol
@@ -60,8 +60,8 @@ contract Flippy {
         uint256 bid,   // [rad]
         uint256 lot,   // [wad]
         address guy,
-        uint48  tic,
-        uint48  end,
+        uint48  tic,   // [unix epoch time]
+        uint48  end,   // [unix epoch time]
         address usr,
         address gal,
         uint256 tab    // [rad]
@@ -200,8 +200,8 @@ contract End is LibNote {
     Spotty   public spot;
 
     uint256  public live;  // Cage flag
-    uint256  public when;  // Time of cage
-    uint256  public wait;  // Processing Cooldown Length
+    uint256  public when;  // Time of cage                   [unix epoch time]
+    uint256  public wait;  // Processing Cooldown Length             [seconds]
     uint256  public debt;  // Total outstanding dai following processing [rad]
 
     mapping (bytes32 => uint256) public tag;  // Cage price              [ray]
--- a/src/end.sol
+++ b/src/end.sol
@@ -199,7 +199,7 @@ contract End is LibNote {
     PotLike  public pot;
     Spotty   public spot;
 
-    uint256  public live;  // Cage flag
+    uint256  public live;  // Active Flag
     uint256  public when;  // Time of cage                   [unix epoch time]
     uint256  public wait;  // Processing Cooldown Length             [seconds]
     uint256  public debt;  // Total outstanding dai following processing [rad]
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -66,7 +66,7 @@ contract Flapper is LibNote {
     uint48   public   ttl = 3 hours;  // 3 hours bid duration         [seconds]
     uint48   public   tau = 2 days;   // 2 days total auction length  [seconds]
     uint256  public kicks = 0;
-    uint256  public live;  // Shutdown Flag
+    uint256  public live;  // Active Flag
 
     // --- Events ---
     event Kick(
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -49,8 +49,8 @@ contract Flapper is LibNote {
 
     // --- Data ---
     struct Bid {
-        uint256 bid;
-        uint256 lot;
+        uint256 bid;  // gems paid     [wad]
+        uint256 lot;  // dai for sale  [rad]
         address guy;  // high bidder
         uint48  tic;  // expiry time
         uint48  end;
@@ -58,7 +58,7 @@ contract Flapper is LibNote {
 
     mapping (uint => Bid) public bids;
 
-    VatLike  public   vat;
+    VatLike  public   vat; // CDP Engine
     GemLike  public   gem;
 
     uint256  constant ONE = 1.00E18;
@@ -66,7 +66,7 @@ contract Flapper is LibNote {
     uint48   public   ttl = 3 hours;  // 3 hours bid duration
     uint48   public   tau = 2 days;   // 2 days total auction length
     uint256  public kicks = 0;
-    uint256  public live;
+    uint256  public live;  // Access Flag
 
     // --- Events ---
     event Kick(
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -30,7 +30,7 @@ contract GemLike {
 /*
    This thing lets you sell some dai in return for gems.
 
- - `lot` dai for sale
+ - `lot` dai in return for bid
  - `bid` gems paid
  - `ttl` single bid lifetime
  - `beg` minimum bid increase
@@ -49,11 +49,11 @@ contract Flapper is LibNote {
 
     // --- Data ---
     struct Bid {
-        uint256 bid;  // gems paid            [wad]
-        uint256 lot;  // dai for sale         [rad]
+        uint256 bid;  // gems paid               [wad]
+        uint256 lot;  // dai in return for bid   [rad]
         address guy;  // high bidder
-        uint48  tic;  // bid expiry time      [unix epoch time]
-        uint48  end;  // auction expiry time  [unix epoch time]
+        uint48  tic;  // bid expiry time         [unix epoch time]
+        uint48  end;  // auction expiry time     [unix epoch time]
     }
 
     mapping (uint => Bid) public bids;
@@ -66,7 +66,7 @@ contract Flapper is LibNote {
     uint48   public   ttl = 3 hours;  // 3 hours bid duration
     uint48   public   tau = 2 days;   // 2 days total auction length
     uint256  public kicks = 0;
-    uint256  public live;  // Access Flag
+    uint256  public live;  // Shutdown Flag
 
     // --- Events ---
     event Kick(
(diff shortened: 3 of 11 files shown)


src/end.sol
```diff
@@ -43,9 +43,9 @@ contract VatLike {
 }
 contract CatLike {
     function ilks(bytes32) external returns (
-        address flip,  // Liquidator
-        uint256 chop,  // Liquidation Penalty   [ray]
-        uint256 lump   // Liquidation Quantity  [rad]
```
brianmcmichael, line 48: Did you mean to remove the comments around these three?

KentonPrescott: Yes

src/cat.sol
```diff
@@ -62,9 +62,9 @@ contract Cat is LibNote {
 
     mapping (bytes32 => Ilk) public ilks;
 
-    uint256 public live;
-    VatLike public vat;
-    VowLike public vow;
+    uint256 public live; // Access Flag
```
kmbarry1, line 65: nit: minimum of two spaces between code and start of comment

src/flap.sol
```diff
@@ -49,24 +49,24 @@ contract Flapper is LibNote {
...
+        uint256 lot;  // dai for sale  [rad]
         address guy;  // high bidder
         uint48  tic;  // expiry time
         uint48  end;
     }
 
     mapping (uint => Bid) public bids;
 
-    VatLike  public   vat;
+    VatLike  public   vat; // CDP Engine
```
kmbarry1, line 61: nit: minimum of two spaces between code and start of comment

kmbarry1: I also sort of question whether this comment really adds much--anyone who wants to understand the system is going to have to learn what the `Vat` is pretty much immediately.

src/flop.sol
```diff
@@ -50,16 +50,16 @@ contract Flopper is LibNote {
...
+        uint256 lot;  // gems for sale [wad]
         address guy;  // high bidder
         uint48  tic;  // expiry time
         uint48  end;
     }
 
     mapping (uint => Bid) public bids;
 
-    VatLike  public   vat;
+    VatLike  public   vat; // CDP Engine
```
kmbarry1, line 62: Same comments as analogous line in `flap.sol`.

src/flop.sol
```diff
@@ -50,16 +50,16 @@ contract Flopper is LibNote {
 
     // --- Data ---
     struct Bid {
-        uint256 bid;
-        uint256 lot;
+        uint256 bid;  // dai paid      [rad]
+        uint256 lot;  // gems for sale [wad]
```
kmbarry1, line 54: I'm not totally satisfied with "gems for sale" here as an accurate description. I'm struggling to come up with a suitably pithy replacement, but maybe consider some of the below options:

"gems bidder will accept"
"gems to receive"
"gems accepted in exchange for bid"
"gems bidder will accept in exchange for bid"
"gems in return"
"gems in return for bid"
"gems in return for bid upon win"
"gems to pay out"

Maybe "gems to receive" works best...

src/jug.sol
```diff
@@ -22,14 +22,14 @@ contract Jug is LibNote {
 
     // --- Data ---
     struct Ilk {
-        uint256 duty;
-        uint256  rho;
+        uint256 duty;  // Stability Fee      [ray]
```
kmbarry1, line 25: Just labelling this "Stability Fee" is dangerous because it really isn't the stability fee, that's `base + duty`, and if both of those values are non-zero, the non-linear correspondence between addition of per-second rates and the effective yearly rate creates a lot of risk of confusion. I know it's more verbose, but something like "collateral-specific per-second stability fee contribution" is likely to be more robustly understood.

src/jug.sol
```diff
@@ -22,14 +22,14 @@ contract Jug is LibNote {
...
+        uint256  rho;  // Time of last drip  [seconds]
     }
 
     mapping (bytes32 => Ilk) public ilks;
-    VatLike                  public vat;
-    address                  public vow;
-    uint256                  public base;
+    VatLike                  public vat;  // CDP Engine
+    address                  public vow;  // Debt Engine
+    uint256                  public base; // Global Stability Fee [wad]
```
kmbarry1, line 32: Similar comment here to above--something like "global per-second stability fee contribution" is less likely to be misinterpreted.

Also, minimum of two spaces between code and start of comment.

src/pot.sol
```diff
@@ -54,17 +54,17 @@ contract Pot is LibNote {
     }
 
     // --- Data ---
-    mapping (address => uint256) public pie;  // user Savings Dai
+    mapping (address => uint256) public pie;  // User Savings Dai [wad]
```
kmbarry1, line 57: So here's a case where I feel the original comment is not really accurate, because the quantity isn't Dai, it's "scaled dai", much like `art` is "scaled debt". So it would be fair, if a little cryptic, to call this "per-user scaled savings dai". More verbose but more clear would be "per-user savings dai divided by chi". I'm fine with either of those options.

src/pot.sol
```diff
@@ -54,17 +54,17 @@ contract Pot is LibNote {
...
 
-    VatLike public vat;  // CDP engine
-    address public vow;  // debt engine
-    uint256 public rho;  // time of last drip
+    VatLike public vat;  // CDP Engine
+    address public vow;  // Debt Engine
+    uint256 public rho;  // Time of last drip     [seconds]
 
-    uint256 public live;  // Access Flag
+    uint256 public live; // Access Flag
```
kmbarry1, line 67: nit: minimum of two spaces between code and start of comment

src/spot.sol
```diff
@@ -37,22 +37,22 @@ contract Spotter is LibNote {
...
-        uint256 mat;
+        PipLike pip;  // OSM
+        uint256 mat;  // Liquidation ratio [ray]
     }
 
     mapping (bytes32 => Ilk) public ilks;
 
-    VatLike public vat;
-    uint256 public par; // ref per dai
+    VatLike public vat; // CDP Engine
```
kmbarry1, line 46: nit: minimum of two spaces between code and start of comment

src/spot.sol
```diff
@@ -37,22 +37,22 @@ contract Spotter is LibNote {
...
+    uint256 public par; // ref per dai     [ray]
```
kmbarry1, line 47: nit: minimum of two spaces between code and start of comment

Also I'm not convinced it's aesthetically more pleasing to align the `[ray]` on this line with the one on line 41, as opposed to just giving it a natural single space.

src/spot.sol
kmbarry1: Here and elsewhere I still question whether there's any point to labelling the `Vat` like this.

src/vow.sol
```diff
@@ -50,22 +50,22 @@ contract Vow is LibNote {
...
+    uint256 public dump;  // Flop initial lot size  [wad]
+    uint256 public sump;  // Flop fixed bid size    [rad]
 
-    uint256 public bump;  // flap fixed lot size    [rad]
-    uint256 public hump;  // surplus buffer       [rad]
+    uint256 public bump;  // Flap fixed lot size    [rad]
+    uint256 public hump;  // Surplus buffer         [rad]
 
-    uint256 public live;
+    uint256 public live;  // Access Flag
```
kmbarry1, line 68: So it took me until the last file to realize this, but I really dislike "Access Flag" as the description for `live` variables. I think "Shutdown Flag" would be much more informative--do you agree?

src/flip.sol
```diff
@@ -50,14 +50,14 @@ contract Flipper is LibNote {
 
     // --- Data ---
     struct Bid {
-        uint256 bid;
-        uint256 lot;
+        uint256 bid;  // dai paid          [rad]
+        uint256 lot;  // gems for sale     [wad]
         address guy;  // high bidder
         uint48  tic;  // expiry time
```
kmbarry1, line 56: `// bid expiry time [unix epoch time]`

src/flip.sol
```diff
@@ -50,14 +50,14 @@ contract Flipper is LibNote {
...
         uint48  end;
```
kmbarry1, line 57: `auction expiry time [unix epoch time]`

src/flop.sol
```diff
@@ -50,16 +50,16 @@ contract Flopper is LibNote {
...
         address guy;  // high bidder
         uint48  tic;  // expiry time
```
kmbarry1, line 56: `tic` & `end` should have comments similar to those suggested for the `Flipper`.

src/flap.sol
```diff
@@ -49,24 +49,24 @@ contract Flapper is LibNote {
 
     // --- Data ---
     struct Bid {
-        uint256 bid;
-        uint256 lot;
+        uint256 bid;  // gems paid     [wad]
+        uint256 lot;  // dai for sale  [rad]
         address guy;  // high bidder
         uint48  tic;  // expiry time
```
kmbarry1, line 55: `tic` & `end` should have comments similar to those suggested below for the `Flipper`

kmbarry1 (commented):
made some suggestions

src/flop.sol
godsflaw: bikeshed: I like "gems in return for bid"

src/jug.sol
godsflaw: I second this, a number of people have already stumbled over this detail.

src/vow.sol
godsflaw: Yes, shutdown flag or caged flag is how I think about this.  Although, `live` is probably the most descriptive variable name in the entire codebase, we may not even need the comment.

godsflaw (changes requested):
I strongly agree with this PR for the typing alone: `[wad]`, `[ray]`, `[rad]`, and `[unix epoch time]`.  This will save integrators loads of time looking these values up on-chain and figuring out their magnitudes.  There are a few suggestions in the comments below.  In `end.sol` for the `Flippy` contract, you should probably add `[unix epoch time]` to `tic` and `end` there.  Thanks for the work @KentonPrescott 👍

src/flap.sol
KentonPrescott: I'm indifferent to adding short descriptions to contracts. The `Pot` originally contained short descriptions for `Vat` and `Vow` as `CDP Engine` and `Debt Engine`, so I simply persisted this pattern across the repo - I did not, however, attempt to re-describe the `Cat`, `Pot`, or `Spotty` in `End`, as they aren't used as much as the former two and would cause require more time to come to consensus on the names. The other variable names, on the other hand, have already been described in the Glossary, so it's easy to use those (or slight) variants and move forward. 

Up to you how you want to proceed - I'd be happy to delete short descriptions for contract.

src/flop.sol
KentonPrescott: "gems for sale" was used from commented auction structure in L27-39. 
However, I can update both this line and the original reference in L31. 

I like "gems to receive" as well but don't mind any other description. Just let me know

src/jug.sol
KentonPrescott: Completely agree. Later on, the glossary should be updated with the description we adopt here. 
I think the benefit in comprehension outweighs the cost of verbosity. 

I like "Collateral-specific, per-second stability fee contribution".

Definitely worth another discussion, but I think `base` should be set to `1` and `duty` is set to the collateral-specific component amount; any system-wide stability fee that's enforced can be contained within `base` (e.g. `1.00000000XYZ`). It's much easier to explain to integration partners (had to add some nuance on this earlier today) and seems to be the intended method of implementing rates.

src/jug.sol
KentonPrescott: 👍 check on my reply with regards to `duty`

src/pot.sol
KentonPrescott: Agreed. Taking inspiration from `art // Normalised Debt`, what about `Normalised Savings Dai` ?

src/spot.sol
KentonPrescott: You're right - will give it a natural single space

src/vow.sol
KentonPrescott: I like Shutdown Flag. When live = 0, we want it to undeniably be understood that shutdown has occurred.

src/flop.sol
kmbarry1: I'm fine with either "gems to receive" or "gems in return for bid". As the PR author, Kenton, the choice is at your discretion when the reviewers are ambivalent :).

src/pot.sol
kmbarry1: Sure, let's go with that.

src/flap.sol
kmbarry1: Eh, it probably isn't worth the time to remove them.

src/vow.sol
kmbarry1: Yeah I would vote for changing to `// Shutdown Flag` globally. Or I guess since `1` means "not shut down", maybe it's really the `// Active Flag`...eh, I think I still like `// Shutdown Flag`. Or no comment is probably fine too.

kmbarry1 (commented):
So I didn't exhaustively check all the integer types, but I think I looked at most of them and they were all correct AFAICT. I think once the remaining open comments are addressed this will be ready to merge.

src/jug.sol
kmbarry1: Go ahead with the verbose description but let's avoid in-depth discussion of what the best convention for using `base` and `duty` is in this PR ;). (There's a not-unreasonable counter-argument that says `base` should be a fractional value and `duty` should be `>= ONE` based on the fact that `init` sets `duty` to `ONE`, but this is no place to bikeshed on that...the meta-point is that only `base+duty` has an unambiguous interpretation. And I'm personally a fan of "`base` is superfluous, shouldn't be used at all, and in fact shouldn't exist in future versions of the system".).

src/flop.sol
KentonPrescott: Going with "gems in return for bid"

src/jug.sol
KentonPrescott: Sounds good. I'll go with "Collateral-specific, per-second stability fee contribution".

I assume you're aware, but for record, it's important that any changes (to how rates are enforced today) are shared with INTG team, so we can push outward to our network 👍 . Thanks again.

src/vow.sol
KentonPrescott: will go with `// Shutdown Flag`

gbalabasquer: Well thinking that 1 indicates that the system is not caged, i'd rather go for the positive naming. So I prefer "Active Flag" than "Shutdown Flag" which would be the correct description for an `off` kind variable.
What about "Liveness Flag"? Would it make sense?

src/jug.sol
```diff
@@ -22,14 +22,14 @@ contract Jug is LibNote {
 
     // --- Data ---
     struct Ilk {
-        uint256 duty;
-        uint256  rho;
+        uint256 duty;  // Stability Fee      [ray]
+        uint256  rho;  // Time of last drip  [seconds]
```
gbalabasquer, line 26: [unix epoch time]

src/jug.sol
```diff
@@ -22,14 +22,14 @@ contract Jug is LibNote {
...
     }
 
     mapping (bytes32 => Ilk) public ilks;
-    VatLike                  public vat;
-    address                  public vow;
-    uint256                  public base;
+    VatLike                  public vat;   // CDP Engine
+    address                  public vow;   // Debt Engine
+    uint256                  public base;  // Global Stability Fee [wad]
```
gbalabasquer, line 32: `base` is also a `ray` as `duty` is (regardless which has the `+1 RAY` value).

src/pot.sol
```diff
@@ -54,15 +54,15 @@ contract Pot is LibNote {
...
+    uint256 public Pie;   // Total Savings Dai     [wad]
+    uint256 public dsr;   // The Dai Savings Rate  [ray]
+    uint256 public chi;   // The Rate Accumulator  [ray]
 
-    VatLike public vat;  // CDP engine
-    address public vow;  // debt engine
-    uint256 public rho;  // time of last drip
+    VatLike public vat;   // CDP Engine
+    address public vow;   // Debt Engine
+    uint256 public rho;   // Time of last drip     [seconds]
```
gbalabasquer, line 65: [unix epoch time]

src/vow.sol
```diff
@@ -50,22 +50,22 @@ contract Vow is LibNote {
...
-    uint256 public Sin;   // queued debt          [rad]
-    uint256 public Ash;   // on-auction debt      [rad]
+    mapping (uint256 => uint256) public sin;  // debt queue
+    uint256 public Sin;   // Queued debt            [rad]
+    uint256 public Ash;   // On-auction debt        [rad]
 
-    uint256 public wait;  // flop delay
-    uint256 public dump;  // flop initial lot size  [wad]
-    uint256 public sump;  // flop fixed bid size    [rad]
+    uint256 public wait;  // Flop delay
```
gbalabasquer, line 61: [seconds] would be appropriate here

KentonPrescott:
@gbalabasquer @godsflaw @kmbarry1 
The second, and hopefully final, round of revisions have been committed. It includes all requested changes up until this point.

src/cat.sol
```diff
@@ -62,9 +62,9 @@ contract Cat is LibNote {
 
     mapping (bytes32 => Ilk) public ilks;
 
-    uint256 public live;
-    VatLike public vat;
-    VowLike public vow;
+    uint256 public live;  // Shutdown Flag
```
gbalabasquer, line 65: Hmm I have to insist that IMO "Shutdown Flag" would lead me to think that 1 means the system is shutdown.
If all the rest still prefers it I'll not oppose, but it's kind of an important param to make people confused about how it is used.

src/end.sol
```diff
@@ -193,24 +193,24 @@ contract End is LibNote {
...
-    VowLike  public vow;
+    VowLike  public vow;   // Debt Engine
     PotLike  public pot;
     Spotty   public spot;
 
-    uint256  public live;  // cage flag
-    uint256  public when;  // time of cage
-    uint256  public wait;  // processing cooldown length
-    uint256  public debt;  // total outstanding dai following processing [rad]
+    uint256  public live;  // Cage flag
```
gbalabasquer, line 202: I guess this one should use the same description than the rest of `live` flags.

src/flap.sol
```diff
@@ -49,24 +49,24 @@ contract Flapper is LibNote {
...
     mapping (uint => Bid) public bids;
 
-    VatLike  public   vat;
+    VatLike  public   vat;  // CDP Engine
     GemLike  public   gem;
 
     uint256  constant ONE = 1.00E18;
     uint256  public   beg = 1.05E18;  // 5% minimum bid increase
     uint48   public   ttl = 3 hours;  // 3 hours bid duration
     uint48   public   tau = 2 days;   // 2 days total auction length
```
gbalabasquer, line 67: Maybe we should indicate here that the variables are saved as `[seconds]`?

src/pot.sol
```diff
@@ -54,17 +54,17 @@ contract Pot is LibNote {
     }
 
     // --- Data ---
-    mapping (address => uint256) public pie;  // user Savings Dai
+    mapping (address => uint256) public pie;  // Normalised Savings Dai [wad]
 
-    uint256 public Pie;  // total Savings Dai
-    uint256 public dsr;  // the Dai Savings Rate
-    uint256 public chi;  // the Rate Accumulator
+    uint256 public Pie;   // Total Savings Dai     [wad]
```
gbalabasquer, line 59: Should we better call it `Total Normalised Savings Dai`?

src/spot.sol
```diff
@@ -37,22 +37,22 @@ contract Spotter is LibNote {
 
     // --- Data ---
     struct Ilk {
-        PipLike pip;
-        uint256 mat;
+        PipLike pip;  // OSM
```
gbalabasquer, line 40: This actually depends on the collateral. In general will be an OSM. But for example for SAI collateral it was just a simple DSValue. So I'd call it a generic `Price Feed`.

src/cat.sol
kmbarry1: Yeah, you have a good point...I don't find "Shutdown Flag" confusing because I'm already familiar with the codebase, but for someone who isn't, it may be a different story.

KentonPrescott: On second thought, I think "Access Flag" is more accurate.

kmbarry1: "Access Flag" is the only description I am categorically opposed to.

I think "Active Flag" or maybe "Operational Flag" or even just "Liveness Flag" are all better than "Access Flag".

No comment is also a better option than "Access Flag".

gbalabasquer: I like "Active Flag" or "Liveness Flag".

KentonPrescott: Going with `Active Flag`

godsflaw (approved):
LGTM  🍏

gbalabasquer (approved):
LGTM. Please use `squash` option when merging.

src/end.sol
```diff
@@ -193,24 +193,24 @@ contract End is LibNote {
...
+    VowLike  public vow;   // Debt Engine
     PotLike  public pot;
     Spotty   public spot;
 
-    uint256  public live;  // cage flag
-    uint256  public when;  // time of cage
-    uint256  public wait;  // processing cooldown length
-    uint256  public debt;  // total outstanding dai following processing [rad]
+    uint256  public live;  // Active Flag
+    uint256  public when;  // Time of cage                   [unix epoch time]
```
kmbarry1, line 203: FYI the way unit labels are aligned is not consistent throughout the PR...here we right-justify, other places we left-justify, and there's even one place that's neither (mid-justified, maybe?). If there is some aesthetic intention in each case then NBD, but thought I'd point it out.

kmbarry1 (approved):
LGTM, +1 to squash-and-merge

PR #114, opened 2020-06-11
Add dec var to ETHJoin and DaiJoin adapters

brianmcmichael:
GemJoin adapters include a public `dec` variable. This adds `dec` variables to the ETHJoin and DaiJoin adapters for interface compatibility.

### DIFF
--- a/src/join.sol
+++ b/src/join.sol
@@ -111,13 +111,15 @@ contract ETHJoin is LibNote {
 
     VatLike public vat;   // CDP Engine
     bytes32 public ilk;   // Collateral Type
-    uint    public live;  // Active Flag
+    uint256 public live;  // Active Flag
+    uint256 public dec;
 
     constructor(address vat_, bytes32 ilk_) public {
         wards[msg.sender] = 1;
         live = 1;
         vat = VatLike(vat_);
         ilk = ilk_;
+        dec = 18;
     }
     function cage() external note auth {
         live = 0;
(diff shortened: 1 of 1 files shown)


gbalabasquer:
Hmmm I'm not really sure about this PR.

Regarding `ETHJoin`, it is not being used, and I think is clear ETH is 18 decimals. It would actually require different proxy actions meaning that the standardisation might not be necessary. Anyway I'm not opposed to add to it.

Regarding `DaiJoin` is a totally different adapter, so I don't really see the need here for standardisation with the collateral ones. On the other hand if we modify it we would probably want to update it in kovan and mainnet as well, which will be a huge headache (and not 100% if possible to do keeping the same DAI token). So I'd prefer to keep as it is so the deployed contracts keep matching with what we have in Github.

gbalabasquer:
Have rebased master for the travis fix here as well.

brianmcmichael:
Totally understandable. I wonder if we should have a `develop` branch where we can include potential updates for future iterations or re-deployments. This is obviously not a crucial addition but it would be nice to have a conforming interface for adapters if we ever did need to redeploy. In this case, having the `dec` var standard across adapters would make programmatic integration of third-party dapps a little easier.

gbalabasquer:
Yes, at some point we need to have to separate the launched version from an updated one. I guess we will like to maintain a solc 0.6.x compatible version as well.

Thugnasty777:
Join

kasperwi:
Access

Thugnasty777:
Access

PR #115, opened 2020-06-11
Upgrade dependencies

brianmcmichael:
Upgrade ds-test and ds-token to latest

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 os:
   - linux
 language: nix
+nix: 2.3.6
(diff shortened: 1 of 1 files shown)


gbalabasquer:
This needs to include upgrading `ds-value` as there is a remapping error now.

gbalabasquer:
I've rebase `master` with the fix of the travis file. It should still fail due the remappings.

brianmcmichael:
Added `ds-value` upgrade

PR #120, opened 2020-07-14
Solc 0.5 0.6

gbalabasquer:


kmbarry1 (commented):
LGTM

Do we have another branch or something that tracks what's on mainnet specifically? If not, might be good to have that...

kmbarry1 (approved):

kmbarry1:
(my previous comment should in no way be considered blocking)

gbalabasquer:
> LGTM
> 
> Do we have another branch or something that tracks what's on mainnet specifically? If not, might be good to have that...

No actually, but the only change regarding what we have today in mainnet are the new flip, flop and flap auctions which soon will be updated. These changes are only modifying interfaces (which should not change anything at bytecode level) and also adding the `()` for a few assembly functions. These `()` are the only ones that I'm not 100% sure if produce different bytecode when compiling.

godsflaw (approved):
LGTM, I also approve of tagging what we have on mainnet just in case we need to debug that specific code.

gbalabasquer:
> LGTM, I also approve of tagging what we have on mainnet just in case we need to debug that specific code.

Let's confirm if the `()` actually modify something of the bytecode, if not the case (which I think) we can wait for the new auctions contracts to be deployed and tag that version.

wilbarnes (approved):

PR #123, opened 2020-08-05
SC-4900: support to limit on-auction DAI

godsflaw:
This PR adds support to limit the amount of on-auction Dai.

Governance sets the variable `box` to be the maximum amount of Dai out for auction at any given time.  As under-collateralized Vaults are sent to auction, an accumulator `litter` is incremented.  Once the `litter` fills the `box`, the `cat` can no longer `bite` thus pausing liquidations.  As Dai is reclaimed in `flip` auctions, `scoop()` is called to decrement the `litter` in the `box`; thus making room for more liquidations.

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -143,9 +143,6 @@ contract Cat is LibNote {
 
     // --- CDP Liquidation ---
     function bite(bytes32 ilk, address urn) external returns (uint id) {
-        // NOTE: because of stack-depth limits, we need to re-use num in this
-        // function.  At the start, num is spot, then later becomes lot.
-
         (, uint rate, uint spot) = vat.ilks(ilk);
         (uint ink, uint art) = vat.urns(ilk, urn);
 
@@ -166,10 +163,8 @@ contract Cat is LibNote {
         // Pick a fractional art that doesn't put us over box
         uint fart = min(
             art, rdiv((sub(box, litter) / ilkS.chop), rate)
-      //WAD=WAD, WAD       RAD, RAD     / RAY            / RAY
         );
         lot = min(lot, wmul(lot, wdiv(fart, art)));
-      //WAD       WAD, WAD
 
         require(lot <= 2**255 && fart <= 2**255, "Cat/overflow");
         vat.grab(ilk, urn, address(this), address(vow), -int(lot), -int(fart));
@@ -177,7 +172,6 @@ contract Cat is LibNote {
         // Accumulate litter in the box
         uint tab = rmul(mul(fart, rate), ilkS.chop);
         litter = add(litter, tab);
-      //RAD          RAD   , RAD  RAD WAD , RAY  , RAY
 
         vow.fess(mul(fart, rate));
         id = Kicker(ilkS.flip).kick({
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -138,7 +138,6 @@ contract Cat is LibNote {
 
         require(live == 1, "Cat/not-live");
         require(spot > 0 && mul(ink, spot) < mul(art, rate), "Cat/not-unsafe");
-        // TODO(cmooney): test me
         require(litter < box, "Cat/liquidation-limit-hit");
 
         Ilk memory milk = ilks[ilk];
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -26,13 +26,13 @@ interface Kicker {
 
 interface VatLike {
     function ilks(bytes32) external view returns (
-        uint256 Art,   // [wad]
-        uint256 rate,  // [ray]
-        uint256 spot   // [ray]
+        uint256 Art,  // [wad]
+        uint256 rate, // [ray]
+        uint256 spot  // [ray]
     );
     function urns(bytes32,address) external view returns (
-        uint256 ink,   // [wad]
-        uint256 art    // [wad]
+        uint256 ink,  // [wad]
+        uint256 art   // [wad]
     );
     function grab(bytes32,address,address,address,int,int) external;
     function hope(address) external;
@@ -56,17 +56,17 @@ contract Cat is LibNote {
     // --- Data ---
     struct Ilk {
         address flip;  // Liquidator
-        uint256 chop;  // Liquidation Penalty   [ray]
-        uint256 lump;  // Liquidation Quantity  [wad]
+        uint256 chop;  // Liquidation Penalty  [ray]
+        uint256 lump;  // Liquidation Quantity [wad]
     }
 
     mapping (bytes32 => Ilk) public ilks;
 
-    uint256 public box;  // Max Dai out for liquidation [rad]
-    uint256 public litter;  // Balance of Dai out for liquidation [rad]
-    uint256 public live;  // Active Flag
-    VatLike public vat;   // CDP Engine
-    VowLike public vow;   // Debt Engine
+    uint256 public box;    // Max Dai out for liquidation        [rad]
+    uint256 public litter; // Balance of Dai out for liquidation [rad]
+    uint256 public live;   // Active Flag
+    VatLike public vat;    // CDP Engine
+    VowLike public vow;    // Debt Engine
 
     // --- Events ---
     event Bite(
@@ -88,10 +88,10 @@ contract Cat is LibNote {
     }
 
     // --- Math ---
+    uint constant MLN = 10 **  6;
     uint constant WAD = 10 ** 18;
     uint constant RAY = 10 ** 27;
     uint constant RAD = 10 ** 45;
-    uint constant MLN = 10 **  6;
 
     function min(uint x, uint y) internal pure returns (uint z) {
         if (x > y) { z = y; } else { z = x; }
@@ -163,15 +163,15 @@ contract Cat is LibNote {
         // art = ----------------------
         //               rate
         //
-        // pick a fractional art that doesn't put us over box
+        // Pick a fractional art that doesn't put us over box
         uint fart = min(
             art, rdiv((sub(box, litter) / ilks[ilk].chop), rate)
       //WAD=WAD, WAD       RAD, RAD     / RAY            / RAY
         );
         num = min(num, wmul(num, wdiv(fart, art)));
-      //WAD     WAD, WAD
+      //WAD       WAD, WAD
 
-        // accumulate litter in the box
+        // Accumulate litter in the box
         litter = add(litter, rmul(mul(fart, rate), ilks[ilk].chop));
       //RAD          RAD   , RAD  RAD WAD , RAY  , RAY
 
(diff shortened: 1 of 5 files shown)


lucas-manuel:
This LGTM, all existing tests are passing. Regarding `num`, I think it is okay as long as we make sure to leave that comment in the contract.

gbalabasquer:
So I think there is an issue with this solution. Each time you run an auction you add what you would expect as maximum DAI recover from the auction to `litter`. However the auction might end in `tend` phase making the accumulator will not be able to reach 0 again.

gbalabasquer:
I think the solution is just substracting the total on first `dent` or in `deal` (if it didn't have a `dent` phase).
Probably easier to always do it in `deal`.

gbalabasquer:
Other option is keep doing the subtraction in each `tend` and do `tab - bids[id]` in first `dent` or `deal`.

src/flip.sol
```diff
@@ -177,6 +188,7 @@ contract Flipper is LibNote {
         require(bids[id].bid < bids[id].tab, "Flipper/already-dent-phase");
         vat.flux(ilk, address(this), msg.sender, bids[id].lot);
         vat.move(msg.sender, bids[id].guy, bids[id].bid);
+        cat.scoop(bids[id].bid);
```
gbalabasquer, line 191: It should be `cat.scoop(bids[id].tab)` here.

src/cat.sol
```diff
@@ -123,22 +148,45 @@ contract Cat is LibNote {
 
         require(live == 1, "Cat/not-live");
         require(spot > 0 && mul(ink, spot) < mul(art, rate), "Cat/not-unsafe");
+        require(litter < box, "Cat/liquidation-limit-hit");
+
+        Ilk memory ilkS = ilks[ilk];
```
lucas-manuel, line 153: Should we change this to `milk` maybe?

src/test/vat.t.sol
```diff
@@ -522,13 +525,29 @@ contract BiteTest is DSTest {
         me = address(this);
     }
 
+    function test_set_lump_multiple_ilks() public {
+        cat.file("gold",   "lump", rad(115792 ether));
+        (,, uint256 goldLump) = cat.ilks("gold");
+        assertEq(goldLump, rad(115792 ether));
+        cat.file("silver", "lump", rad(115792 ether));
+        (,, uint256 silverLump) = cat.ilks("gold");
```
lucas-manuel, line 533: This should be `(,, uint256 silverLump) = cat.ilks("silver");`

lucas-manuel: We should also make different `lump`s for the different ilks so we can assert that they're different

src/cat.sol
```diff
@@ -56,15 +56,17 @@ contract Cat is LibNote {
...
+        uint256 chop;  // Liquidation Penalty  [ray]
+        uint256 lump;  // Liquidation Quantity [rad]
     }
 
     mapping (bytes32 => Ilk) public ilks;
 
-    uint256 public live;  // Active Flag
-    VatLike public vat;   // CDP Engine
-    VowLike public vow;   // Debt Engine
+    uint256 public box;    // Max Dai out for liquidation        [rad]
```
rainbreak, line 65: You might want to preserve the original storage slot ordering. It was just a coincidence that it was alphabetical order too.

src/test/vat.t.sol
godsflaw: good catch

src/cat.sol
godsflaw: good point, tooling could be depending on that.

godsflaw:
> So I think there is an issue with this solution. Each time you run an auction you add what you would expect as maximum DAI recover from the auction to `litter`. However the auction might end in `tend` phase making the accumulator will not be able to reach 0 again.

Moved `cat.scoop()` calls to the `flip.deal()` and `flip.yank()`.  Any better 4-letter meme for `scoop()`?

src/cat.sol
```diff
@@ -81,30 +83,43 @@ contract Cat is LibNote {
     constructor(address vat_) public {
         wards[msg.sender] = 1;
         vat = VatLike(vat_);
+        box = 10 * MLN * RAD;
```
gbalabasquer, line 86: I wouldn't set any default value. We were not doing it for `lump`.

src/cat.sol
```diff
@@ -81,30 +83,43 @@ contract Cat is LibNote {
...
         live = 1;
     }
 
     // --- Math ---
-    uint constant ONE = 10 ** 27;
+    uint256 constant MLN = 10 **  6;
```
gbalabasquer, line 91: If we are not setting a default value for `box`, we can remove it.

src/cat.sol
```diff
@@ -81,30 +83,43 @@ contract Cat is LibNote {
...
+    uint256 constant WAD = 10 ** 18;
```
gbalabasquer, line 92: This constant is not being used. It can be removed.

src/cat.sol
```diff
@@ -81,30 +83,43 @@ contract Cat is LibNote {
...
+    uint256 constant RAY = 10 ** 27;
+    uint256 constant RAD = 10 ** 45;
```
gbalabasquer, line 94: If we are not setting a default value for box, we can remove it.

godsflaw:
ping @lucas-manuel @gbalabasquer @rainbreak @nanexcool @kmbarry1 @andy8052 @brianmcmichael @wilbarnes @livnev 

This is the final version of the code.  It could use one last code review if any of you have the time.  The comment blocks are just there to help in review, and will be removed from the final version.

src/flip.sol
```diff
@@ -57,13 +61,14 @@ contract Flipper is LibNote {
...
-        uint256 tab;  // total dai wanted    [rad]
+        uint256 tab;  // total dai wanted         [rad]
     }
 
-    mapping (uint => Bid) public bids;
+    mapping (uint256 => Bid) public bids;
 
-    VatLike public   vat;
-    bytes32 public   ilk;
+    CatLike public   cat;   // cat liquidation module
```
godsflaw, line 69: Going to preserve storage layout here too, putting the `cat` after `ilk`.

src/cat.sol
```diff
@@ -117,28 +128,42 @@ contract Cat is LibNote {
...
+        (uint256 ink, uint256 art) = vat.urns(ilk, urn);
 
         require(live == 1, "Cat/not-live");
         require(spot > 0 && mul(ink, spot) < mul(art, rate), "Cat/not-unsafe");
+        require(litter < box, "Cat/liquidation-limit-hit");
+
+        Ilk memory milk = ilks[ilk];
+
+        uint256 limit = min(milk.lump, sub(box, litter));
+        uint256 fart = min(art, mul(limit, RAY) / rate / milk.chop);
```
rainbreak, line 142: imo you shouldn't divide by `chop` here as it represents an additional penalty factor on top of the debt that is being covered, rather than any additional debt. `chop` is more of an auction implementation detail and actually was nearly moved to the flipper.

godsflaw: I considered this, but since we're trying to limit the amount of Dai out for liquidation and the `chop` is technically part of that Dai out for liquidation.  In theory, at 13% across multiple collateral types and a 10 million box or larger we could be looking at an additional 1.3 million or more of Dai demand for liquidations.  I think, if this is a consideration that just breaks our theoretical model, that we should still try to capture this Dai in the `box`.

Do you see this as a fundamental break in the accounting or as a break in the theoretical model if how we apply `chop`?  I'm trying to figure out if we should claw this back and perhaps leave the `chop` out of litterbox accounting, or not.  For now, if this just breaks our mental model of `chop` as a feature of auctions, I think we should account for it in the litterbox.  As always, interested in your thoughts.

rainbreak: `chop` only increases the threshold at which the two-way auction flips and begins to return collateral to the cdp owner. Whether the auction reaches that threshold or not depends on the market price. Similarly, the `lot` doesn't exactly represent the "collateral out to auction" as in the reverse auction this amount is reduced. 

Yes, I think `chop` should be left out of the `box`. The `box` would then represent exactly the total cdp debt being covered by liquidation, and it is easy for governance to set it to $10m when they know that really they want the maximum to be $13m, knowing that the penalty is 13%. The accounting isn't really broken otherwise, although I'm wondering about rounding errors, but it doesn't feel great to me that `chop` is part of the calculation for an argument to `grab`.

rainbreak: `claw` is a great meme btw :p, maybe even more apt than `scoop`

src/flip.sol
```diff
@@ -57,19 +61,20 @@ contract Flipper is LibNote {
...
-        uint256 tab;  // total dai wanted    [rad]
+        uint256 tab;  // total dai wanted         [rad]
     }
 
-    mapping (uint => Bid) public bids;
+    mapping (uint256 => Bid) public bids;
 
-    VatLike public   vat;
-    bytes32 public   ilk;
+    VatLike public   vat;            // vat core accounting
```
KentonPrescott, line 69: For consistency with other contracts, such as the `Spotter`, `Jug`, and `End`, consider changing this to `// CDP Engine`

src/cat.sol
kmbarry1: @rainbreak the original motivation for including `chop` in `box` was so that when the auction concludes and `claw` is called, the code doesn't have to correct for `chop` (the auction only knows `art * rate * chop`), which might have changed since the auction began, and would itself be subject to rounding error.

I'm revisiting this design decision right now since it would significantly simplify `bite` and probably governance's reasoning about `box` to leave `chop` out (and reduce precision loss, although I don't think that's a big issue with the current design). The trade-off is that in order to deal with the problem I mentioned above (`chop` can change and you can't perfectly back-calculate even if it hasn't), it's necessary to store the pre-`chop` `tab` value somewhere, either in the `Cat` (indexed by auction id) or in the `Flipper`. Any opinions about that trade-off?

rainbreak: Another option is to move `chop` to the flipper. Or if that's too drastic a change, add a `chop` argument to `kick`, along with the `tab`.

kmbarry1: I thought about putting `chop` in the Flipper (or passing it), but I like keeping the concerns separated--the Flipper just auctions collateral, and the Cat worries about liquidation details like the penalty. I also like not making changes to the Flipper's public APIs (even modest ones) since it's a focus point for integrations.

I have a PR for this over here if you're interested: https://github.com/makerdao/dss/pull/128

src/flip.sol
godsflaw: Changed here: https://github.com/makerdao/dss/pull/124

PR #124, opened 2020-08-17
SC-5246: Further Improvements for LIQ-1.2

godsflaw:
- [x] change meme names
- [x] prevent dusty Vault remainders (won't fix)
- [x] prevent dusty auctions
- [x] consider @rainbreak's arguments against including `chop` in the `litterbox`: https://github.com/makerdao/dss/pull/123#pullrequestreview-466283864
- [x] chop becomes a wad?
- [x] more tests
  - [x] make sure liquidation space left in litterbox is not dusty
  - [x] that we prevent Vaults from being left dusty (won't fix)
  - [x] change all chop tests from ray to wad?
  - [x] `chop` should be set to something non-zero and tests adjusted accordingly

This PR contains further improvements on: https://github.com/makerdao/dss/pull/123

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -152,6 +152,7 @@ contract Cat is LibNote {
 
             dart = min(art, mul(min(milk.dunk, room), RAY) / rate / milk.chop);
         }
+        require(dart > 0, "Cat/null-auction");
 
         uint256 dink = min(ink, mul(ink, dart) / art);
 
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -145,11 +145,11 @@ contract Cat is LibNote {
 
             dart = min(art, mul(min(milk.dunk, room), WAD) / rate / milk.chop);
         }
-        require(dart > 0, "Cat/null-auction");
 
         uint256 dink = min(ink, mul(ink, dart) / art);
 
-        require(dink <= 2**255 && dart <= 2**255, "Cat/overflow");
+        require(dart >  0      && dink >  0     , "Cat/null-auction");
+        require(dart <= 2**255 && dink <= 2**255, "Cat/overflow"    );
 
         // This may leave the CDP in a dusty state
         vat.grab(
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -59,7 +59,7 @@ contract Cat is LibNote {
     struct Ilk {
         address flip;  // Liquidator
         uint256 chop;  // Liquidation Penalty  [ray]
-        uint256 lump;  // Liquidation Quantity [rad]
+        uint256 dunk;  // Liquidation Quantity [rad]
     }
 
     mapping (bytes32 => Ilk) public ilks;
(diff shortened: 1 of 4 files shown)


src/flip.sol
```diff
@@ -25,7 +25,7 @@ interface VatLike {
 }
 
 interface CatLike {
-    function scoop(uint256) external;
+    function claw(uint256) external;
```
rainbreak, line 28: :ok_hand:

src/cat.sol
```diff
@@ -129,41 +136,56 @@ contract Cat is LibNote {
...
             });
         }
 
-        emit Bite(ilk, urn, fink, fart, mul(fart, rate), milk.flip, id);
+        emit Bite(ilk, urn, dink, dart, mul(dart, rate), milk.flip, id);
     }
 
-    function scoop(uint256 poop) external note auth {
-        litter = sub(litter, poop);
+    function claw(uint256 clump) external note auth {
```
rainbreak, line 187: `clump` maybe clearer as `rad`. This would be consistent with the rest of the code where you have a small function with a numeric arg.

kmbarry1 (approved):
LGTM

Assuming there are no bugs, the only work left is expanded testing, so IMO let's merge this so integration work can begin.

There's some risk of needing to redo that work somewhat if we find a bug, but it wouldn't be all of it, and I think it's a risk worth taking.

lucas-manuel (approved):
I'm good with merging this as well, going to branch off of this and add more tests

src/test/vat.t.sol
```diff
@@ -672,18 +633,21 @@ contract BiteTest is DSTest {
 
         // tag=4, mat=2
         vat.file("gold", 'spot', ray(1 ether));  // now unsafe
+        // cat.file("gold", "chop", 1.1 ether);
```
lucas-manuel, line 636: Why is this commented out?

src/test/vat.t.sol
```diff
@@ -672,18 +633,21 @@ contract BiteTest is DSTest {
...
         cat.file("box", rad(75 ether));
-        cat.file("gold", "lump", rad(100 ether));
+        cat.file("gold", "dunk", rad(100 ether));
         assertEq(cat.box(), rad(75 ether));
         assertEq(cat.litter(), 0);
         uint auction = cat.bite("gold", address(this));
-        assertEq(cat.litter(), rad(75 ether));
+        // (,,,,,,,uint tab) = flip.bids(auction);
+        // assertTrue(rad(75 ether) - cat.litter() < ray(3 ether));
+        // assertEq(cat.litter(), tab);
```
lucas-manuel, line 650: Why is this commented out?

godsflaw: These are the tests that we started playing with where we set chop to a non-zero percentage.  This is how we found the dusty remainder, and eventually the dusty auctions, but it was not directly responsible for testing for dust.

I should just remove these comments, but perhaps there is still some desire to make a realistic value tests with a 10% `chop`?

PR #125, opened 2020-08-18
Add codecov and build statuses

zdumitru:


kmbarry1:
Looks like Travis failed due to a request to https://cachix.org/api/v1/cache/maker erroring. Visiting that link seems to work for me in a browser, maybe it was just a transient issue?

brianmcmichael:
I'm ok with this. I will need to learn what these flags do.

PR #130, opened 2020-08-21
SC-5246-null-dink: catch null dink

godsflaw:
This is a contrived example where we end up with a null `dink`.  Since the system can get in this state, we should probably guard against it.  We might want a more real-world example test too.

We should also think through the implications of not being able to liquidate `1 wei` of collateral.

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -148,8 +148,8 @@ contract Cat is LibNote {
 
         uint256 dink = min(ink, mul(ink, dart) / art);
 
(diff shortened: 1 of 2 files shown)


lucas-manuel (commented):
Approved pending the collapse into one `require`

lucas-manuel (approved):

kmbarry1 (approved):
LGTM

lucas-manuel (approved):
LGTM

PR #132, opened 2020-08-24
feat: Add Partial Litterbox Test with Realistic Chop

lucas-manuel:
Setting chop to 13% adds some rounding considerations.

### DIFF
--- a/src/test/vat.t.sol
+++ b/src/test/vat.t.sol
@@ -25,9 +25,6 @@ contract TestVat is Vat {
         dai[usr] += wad * ONE;
         debt += wad * ONE;
     }
-    function balanceOf(address usr) public view returns (uint) {
-        return dai[usr] / ONE;
-    }
 }
 
 contract TestVow is Vow {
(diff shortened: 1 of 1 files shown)


src/test/vat.t.sol
```diff
@@ -676,6 +676,63 @@ contract BiteTest is DSTest {
...
+        uint256 dart = rad(75 ether) * WAD / RAY / 1.13 ether; // room / rate / chop
+        uint256 dink = 100 ether * dart / 150 ether;
+
+        assertEq(ink("gold", address(this)), 100 ether - dink); // Taken in vat.grab
+        assertEq(art("gold", address(this)), 150 ether - dart); // Taken in vat.grab
+        assertEq(vow.sin(now), rad(dart));                      // dart * rate
+        assertEq(gem("gold", address(this)), 900 ether);
+
+        assertEq(vat.balanceOf(address(this)), 150 ether);
+        assertEq(vat.balanceOf(address(vow)),    0 ether);
```
kmbarry1, line 714: Here and elsewhere: why `vat.balanceOf` instead of just `vat.dai`? The latter is arguably more immediately clear in its intent.

kmbarry1: I guess you aren't the one who started this convention, but I personally just find it annoying to use a wrapper for testing when there's a perfectly good getter already.

lucas-manuel: Yeah I was following the convention which is converting `vat.dai` to a wad: 
```
contract TestVat is Vat {
    uint256 constant ONE = 10 ** 27;
    function mint(address usr, uint wad) public {
        dai[usr] += wad * ONE;
        debt += wad * ONE;
    }
    function balanceOf(address usr) public view returns (uint) {
        return dai[usr] / ONE;
    }
}
```
But for this test especially we should keep it in rad. Good catch I'll switch it to `vat.dai`

src/test/vat.t.sol
```diff
@@ -645,35 +644,90 @@ contract BiteTest is DSTest {
...
         assertEq(cat.litter(), 0);
         assertEq(gem("gold", address(this)),  950 ether);
-        assertEq(vat.balanceOf(address(this)), 75 ether);
-        assertEq(vat.balanceOf(address(vow)),  75 ether);
+        assertEq(vat.dai(address(this)),   rad(75 ether));
+        assertEq(vat.dai(address(vow)),    rad(75 ether));
+    }
+
+    // tests a partial lot liquidation because it would fill the literbox
+    function test_partial_litterbox_realistic_values() public {
```
kmbarry1, line 677: Hmmmm if we're being really realistic, we should also have a test with non-unity `rate` as well as `chop`...although I think I do that in at least one `testFail`, so maybe it's technically covered.

lucas-manuel: Added a `rate`

src/test/vat.t.sol
```diff
@@ -645,35 +644,101 @@ contract BiteTest is DSTest {
...
+        assertEq(vow.Woe(), 0 ether);
+        assertEq(gem("gold", address(this)), 900 ether);
+
+        // To check this yourself, use the following rate calculation (example 8%):
+        //
+        // $ bc -l <<< 'scale=27; e( l(1.08)/(60 * 60 * 24 * 365) )'
+        uint256 EIGHT_PCT = 1000000002440418608258400030;
+        jug.file("gold", "duty", EIGHT_PCT);
+        hevm.warp(now + 10 days);
+        jug.drip("gold");
```
kmbarry1, line 698: Just FYI, the easy way to get a non-zero `rate` in a test is to call `Vat.fold` directly since you have `auth` superpowers as the test contract, but it's fine to do it this way.

kmbarry1 (approved):
LGTM

gbalabasquer:
LGTM, please squash on merge.

godsflaw (approved):

PR #133, opened 2020-08-26
View only scratch function to expose liquidation penalty

yaronvel:
I would kindly ask that the cat would expose the tab calculation as a public function.
A similar function exists in compound as an API (in their smart contract).

The motivation is that we are working on an opt-in liquidation system that needs to read the liquidation penalty value from makerdao’s `cat`.
We plan to read the cat address from the `end` contract, but changes like in this version, where the `chop` resolution changed from `rad` to `wad` will break our system (in the future, it is still not deployed).
Hence, a function that for a given `art` returns the tab, would really help.

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -160,7 +160,7 @@ contract Cat is LibNote {
         { // Avoid stack too deep
             // This calcuation will overflow if dart*rate exceeds ~10^14,
             // i.e. the maximum dunk is roughly 100 trillion DAI.
-            uint256 tab = mul(mul(dart, rate), milk.chop) / WAD;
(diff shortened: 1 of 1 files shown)


gbalabasquer:
`rate` and `chop` are public values, that can be read from external contracts and do the math there directly

yaronvel:
@gbalabasquer it is true that they are public. but `chop` is about to change from ray to wad resolution.
Which means it will not always enough to just read it.

Having it as an "official" api, like all other lending platform offer, would make it easier to integrate.

kmbarry1:
Feature request addressed by: https://github.com/makerdao/dss/pull/155

PR #136, opened 2020-09-08
fix: Remove redundant now in flog test

lucas-manuel:
In `wait` test for `flog` the `hevm.warp()` call was warping to `now + tic + 100` when it should have just been `tic + 100`.

### DIFF
--- a/src/test/vow.t.sol
+++ b/src/test/vow.t.sol
@@ -133,7 +133,7 @@ contract VowTest is DSTest {
         uint tic = now;
         vow.fess(100 ether);
         assertTrue(!try_flog(tic) );
-        hevm.warp(now + tic + 100 seconds);
+        hevm.warp(tic + 100 seconds);
         assertTrue( try_flog(tic) );
     }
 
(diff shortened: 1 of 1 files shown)


gbalabasquer (approved):
LGTM.
IMO it wasn't actually redundant but kind of an error. As instead of testing just in the limit when `wait` passes is doing it much further in the future. Still passed, but probably wasn't the idea behind the test.

kmbarry1 (approved):
LGTM

Since `warp` wasn't called previously in that test, I'm pretty sure `now` was just zero so it wasn't affecting anything (unless maybe `warp` is called in `setUp`, I didn't check). But still more correct to remove it, of course.

gbalabasquer:
Yep it is called in the set up: https://github.com/makerdao/dss/blob/update-flog-test/src/test/vow.t.sol#L32.

rainbreak:
Yes, the test wasn't especially precise before, this would have been clearer:

```
        uint tic = now;                                                                                                                                                       
        vow.fess(100 ether);                                                     
        hevm.warp(tic + 99 seconds);                                             
        assertTrue(!try_flog(tic) );                                             
        hevm.warp(tic + 100 seconds);                                            
        assertTrue( try_flog(tic) );  
```

lucas-manuel:
Added @rainbreak's suggestion.

PR #139, opened 2020-09-30
Preliminary Circuit Breaker

andy8052:
Added logic to `stop` and `start` the clipper. All major functions are frozen when `stopped == true`. I see this as step 1 of a 2 step process. The circuit breaker just stops all auctions. Then we can add more functionality to `yank` those auctions when `stopped==true`.

I added a couple tests. All tests are passing.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -81,6 +81,8 @@ contract Clipper {
 
     uint256 internal locked;
 
+    bool stopped;
+
     // --- Events ---
     event Kick(
         uint256  id,
@@ -113,6 +115,11 @@ contract Clipper {
         locked = 1;
         _;
         locked = 0;
+    }
+
+    modifier isStopped(bool which) {
+        require(stopped == which, "Clipper/stopped-incorrect");
+        _;
     }
 
     // --- Administration ---
@@ -156,7 +163,7 @@ contract Clipper {
     function kick(uint256 tab,  // Debt             [rad]
                   uint256 lot,  // Collateral       [wad]
                   address usr   // Liquidated CDP
-    ) external auth returns (uint256 id) {
+    ) external auth isStopped(false) returns (uint256 id) {
         require(kicks < uint256(-1), "Clipper/overflow");
         id = ++kicks;
         active.push(id);
@@ -180,7 +187,7 @@ contract Clipper {
     }
 
     // Reset an auction
-    function warm(uint256 id) external {
+    function warm(uint256 id) isStopped(false) external {
         // Read auction data
         Sale memory sale = sales[id];
         require(sale.tab > 0, "Clipper/not-running-auction");
@@ -209,7 +216,7 @@ contract Clipper {
                   uint256 pay,          // Bid price (DAI / ETH)                            [ray]
                   address who,          // Who will receive the collateral and pay the debt
                   bytes calldata data   
-    ) external lock {
+    ) external lock isStopped(false) {
         // Read auction data
         Sale memory sale = sales[id];
         require(sale.tab > 0, "Clipper/not-running-auction");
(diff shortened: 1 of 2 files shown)


src/test/clip.t.sol
```diff
@@ -623,4 +623,51 @@ contract DutchClipperTest is DSTest {
...
+        assertEq(vat.gem(ilk, me), 960 ether);
+        (ink, art) = vat.urns(ilk, me);
+        assertEq(ink, 40 ether);
+        assertEq(art, 100 ether);
+
+        clip.stop();
+
+        dog.bark(ilk, me);
+    }
+
```
kmbarry1, line 672: Should add coverage for `warm` and `take` as well.

src/clip.sol
```diff
@@ -297,6 +304,14 @@ contract Clipper {
...
     // --- Shutdown ---
 
+    function stop() external auth isStopped(false) {
+        stopped = true;
+    }
+
+    function start() external auth isStopped(true) {
+        stopped = false;
+    }
+
```
kmbarry1, line 314: I'm debating with myself whether these need `isStopped` modifiers... I guess it could help with a caller making a typo-like error (e.g. calling `stop` when the meant `start`), but it could also cause an unexpected revert if someone expects `stop` and `start` calls to be idempotent. For comparison, `ds-stop` doesn't have this check:
https://github.com/dapphub/ds-stop/blob/master/src/stop.sol#L30

kmbarry1: Any considerations I'm missing here?

src/test/clip.t.sol
```diff
@@ -623,4 +623,51 @@ contract DutchClipperTest is DSTest {
         uint256 lotReturn = 30 ether - (rad(60 ether) / ray(4 ether));       // lot - loaf.tab / max = 15
         assertEq(vat.gem(ilk, me), 960 ether + lotReturn);                   // Collateral returned (10 WAD)
     } 
+
+    function test_stop() public {
+        clip.stop();
```
kmbarry1, line 628: Maybe assert on the storage value changing here?

src/test/clip.t.sol
```diff
@@ -623,4 +623,51 @@ contract DutchClipperTest is DSTest {
...
+    }
+
+    function testFail_stop() public {
+        clip.stop();
+        clip.stop();
+    }
+
+    function test_start() public {
+        clip.stop();
+        clip.start();
```
kmbarry1, line 638: ditto

kmbarry1:
Looks like the build failed due to some unrelated tests failing...it seems that it might be related to `hevm` behavior changes. I'll file a ticket for us to look into that this week.

src/test/clip.t.sol
andy8052: I noticed we don't have any coverage for warm. Will add that as well

src/clip.sol
andy8052: Yeah I do not feel strongly here

gbalabasquer: I'd remove the modifiers in these functions.

gbalabasquer:
Not really sure what was discussed regarding the circuit breaker but is it the intent to have a more aggressive circuit breaker than we have now? Otherwise we should just block `bark`, right?

andy8052:
> Not really sure what was discussed regarding the circuit breaker but is it the intent to have a more aggressive circuit breaker than we have now? Otherwise we should just block `bark`, right?

Yeah that was my understanding. I could even dumb down the circuit breaker. Maybe make levels instead of true and false

kmbarry1:
> > Not really sure what was discussed regarding the circuit breaker but is it the intent to have a more aggressive circuit breaker than we have now? Otherwise we should just block `bark`, right?
> 
> Yeah that was my understanding. I could even dumb down the circuit breaker. Maybe make levels instead of true and false

I like the idea of having levels.

PR #140, opened 2020-09-30
Hole per ilk

gbalabasquer:
It doesn't test specifically that `digs` decrements correctly both `dirt` values. But wasn't testing that before either.
In order to check this, we need to use the `takeSetup` modifier, but it would be nice to simplify this modifier to just do the simple set up without any `bark` call inside it.
I guess this can be done later in another PR.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -30,7 +30,7 @@ interface SpotLike {
 }
 
 interface DogLike {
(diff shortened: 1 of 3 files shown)


gbalabasquer:
Closed in favor of: https://github.com/makerdao/dss/pull/141

PR #141, opened 2020-09-30
hole per ilk

gbalabasquer:
It doesn't test specifically that digs decrements correctly both dirt values. But wasn't testing that before either.
In order to check this, we need to use the takeSetup modifier, but it would be nice to simplify this modifier to just do the simple set up without any bark call inside it.
I guess this can be done later in another PR.

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -58,6 +58,8 @@ contract Dog /* is LibNote */ {
     struct Ilk {
         address clip;  // Liquidator
         uint256 chop;  // Liquidation Penalty  [wad]
+        uint256 hole;  // Max DAI needed to cover debt+fees of active auctions per ilk [rad]
+        uint256 dirt;  // Amt DAI needed to cover debt+fees of active auctions per ilk [rad]
     }
 
     mapping (bytes32 => Ilk) public ilks;
@@ -65,8 +67,8 @@ contract Dog /* is LibNote */ {
     uint256 public live;  // Active Flag
     VatLike public vat;   // CDP Engine
     VowLike public vow;   // Debt Engine
-    uint256 public hole;  // Max DAI needed to cover debt+fees of active auctions [rad]
-    uint256 public dirt;  // Amt DAI needed to cover debt+fees of active auctions [rad]
+    uint256 public Hole;  // Max DAI needed to cover debt+fees of active auctions [rad]
+    uint256 public Dirt;  // Amt DAI needed to cover debt+fees of active auctions [rad]
 
     // --- Events ---
     event Bark(
@@ -108,11 +110,12 @@ contract Dog /* is LibNote */ {
         else revert("Dog/file-unrecognized-param");
     }
     function file(bytes32 what, uint256 data) external /* note */ auth {
-        if (what == "hole") hole = data;
+        if (what == "Hole") Hole = data;
         else revert("Dog/file-unrecognized-param");
     }
     function file(bytes32 ilk, bytes32 what, uint256 data) external /* note */ auth {
         if (what == "chop") ilks[ilk].chop = data;
+        else if (what == "hole") ilks[ilk].hole = data;
         else revert("Dog/file-unrecognized-param");
     }
     function file(bytes32 ilk, bytes32 what, address clip) external /* note */ auth {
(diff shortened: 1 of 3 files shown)


gbalabasquer:
I'm not really sure why the CI is failing in the permit tests. It should be pinging to the same hevm version than the `master` branch, which was still passing at least until the last build.
Anyway, it is not related to this specific change, that can be analysed in another PR probably even against to `master`.

kmbarry1:
@gbalabasquer just fyi I have a ticket open to track the build failure:
https://app.clubhouse.io/maker/story/6373/fix-dss-test-failures

lucas-manuel (approved):

src/dog.sol
```diff
@@ -134,9 +137,9 @@ contract Dog /* is LibNote */ {
             (,rate, spot,, dust) = vat.ilks(ilk);
             require(spot > 0 && mul(ink, spot) < mul(art, rate), "Dog/not-unsafe");
 
-            uint256 room = sub(hole, dirt);
+            uint256 room = min(sub(Hole, Dirt), sub(milk.hole, milk.dirt));
 
-            // Test whether the remaining space in the hole is dusty
+            // Test whether the remaining space in the Hole is dusty
```
kmbarry1, line 142: This comment really isn't accurate since it's either the remainder of `Hole` or the per-ilk `hole`. IMO the intent of the code is clear and the comment could actually be removed, but if we keep it, it should be made fully correct.

src/test/clip.t.sol
```diff
@@ -169,55 +145,44 @@ contract DutchClipperTest is DSTest {
 
         me = address(this);
 
-        gov = new DSToken('GOV');
-        gov.mint(100 ether);
-
-        vat = new TestVat();
+        vat = new Vat();
         vat = vat;
```
kmbarry1, line 149: Am I missing something or can this line be removed?

src/test/clip.t.sol
```diff
@@ -473,6 +438,109 @@ contract DutchClipperTest is DSTest {
...
+        (,uint256 chop,, uint256 dirt) = dog.ilks(ilk);
+        assertEq(dirt, 0);
+
+        dog.bark(ilk, me);
+
+        (, uint256 tab, uint256 lot,,,) = clip.sales(1);
+
+        (, uint256 rate,,,) = vat.ilks(ilk);
+
+        assertEq(lot, 40 ether * (tab * 1 ether / rate / chop) / 100 ether);
```
kmbarry1, line 506: I'd used `WAD` here instead of `1 ether`, as is done on line 510.

src/test/clip.t.sol
```diff
@@ -473,6 +438,109 @@ contract DutchClipperTest is DSTest {
...
```
kmbarry1, line 533: Ditto here--`WAD` instead of `1 ether`.

src/test/clip.t.sol
```diff
@@ -473,6 +438,109 @@ contract DutchClipperTest is DSTest {
...
+        assertEq(tab, rad(75 ether) - 0.2 * 10 ** 27); // 2* 10 ** 26 rounding error
```
kmbarry1, line 534: Inconsistent with line 507--`ray(0.2 ether)` versus `0.2 * 10 ** 27`. I prefer either `ray(0.2 ether)` or `0.2 * RAY`.

src/dog.sol
gbalabasquer: Changed a bit the comments. Please let me know if fine now or feel free to tweak them.

src/test/clip.t.sol
gbalabasquer: Yeah, hadn't realized it was there.

src/test/clip.t.sol
gbalabasquer: Done!

src/test/clip.t.sol
gbalabasquer: Done!

src/test/clip.t.sol
gbalabasquer: Fixed, had changed that after but forgot about the other case.

lucas-manuel (approved):
Made a really small update to a comment. Just did another deep dive on the tests and this LGTM.

PR #142, opened 2020-10-01
Use the current price instead of the provided maximum

kmbarry1:
Callers can achieve this anyway using a proxy, and no one wants to pay a higher price than necessary, so they'll all want to do it; therefore, we should just build it into the contract for efficiency.


kmbarry1:
Oops, wrong target branch...will fix.

gbalabasquer:
Target branch can be edited!

PR #143, opened 2020-10-01
Use the current price instead of the provided maximum

kmbarry1:
Callers can achieve this anyway using a proxy, and no one wants to pay a higher price than necessary, so they'll all want to do it; therefore, we should just build it into the contract for efficiency.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -206,7 +206,7 @@ contract Clipper {
     // Buy amt of collateral from auction indexed by id
     function take(uint256 id,           // Auction id
                   uint256 amt,          // Upper limit on amount of collateral to buy       [wad]
-                  uint256 pay,          // Bid price (DAI / ETH)                            [ray]
+                  uint256 max,          // maximum acceptable price (DAI / ETH)             [ray]
                   address who,          // Who will receive the collateral and pay the debt
                   bytes calldata data   
     ) external lock {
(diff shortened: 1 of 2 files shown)


KentonPrescott:
With this change, it seems like `max` (formerly `pay`) is purposed solely for setting expectations for the auction participant. They will either pay the price they request, pay a better price, or the transaction reverts. 

Is there any other use to `max`?

lucas-manuel (approved):
LGTM, tests pass

lucas-manuel:
> With this change, it seems like `max` (formerly `pay`) is purposed solely for setting expectations of the auction participant. They will either pay the price they request, pay a better price, or the transaction reverts.
> 
> Is there any other use to `max`?

Yes, this is correct. We want to make sure that the keepers aren't ever paying more than the maximum price that they specify.

gbalabasquer:
> With this change, it seems like `max` (formerly `pay`) is purposed solely for setting expectations of the auction participant. They will either pay the price they request, pay a better price, or the transaction reverts.
> 
> Does `max` have any other use case?

Actually is a protection in case the auction is reset before the `take` from the keeper is mined.

gbalabasquer:
@lucas-manuel oops sorry I click the wrong place and ask you again for a review.

gbalabasquer (approved):
LGTM

wilbarnes (approved):

kmbarry1:
@KentonPrescott generalizing Gonazlo's point, encapsulating the full economic intent in the calldata is a protection against eclipse attacks (the current auctions do the same thing for the same reason):

https://www.radixdlt.com/post/what-is-an-eclipse-attack/

This prevents and eclipse attacker from broadcasting and eclipse victim's TX out of context.

PR #146, opened 2020-10-04
Clipper cleanup

kmbarry1:
Does a few things:
 - cleans up comments
 - renames `warm` to `redo`
 - adds a `Take` event (Although...do we really need this? `tend` and `dent` don't have events, and it seems simple enough to just watch the storage slots of active auctions to detect changes.)

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -89,6 +89,13 @@ contract Clipper {
         address indexed usr
     );
 
+    event Take(
+        uint256  id,
+        uint256 tab,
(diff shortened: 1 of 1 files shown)


lucas-manuel (approved):
LGTM

gbalabasquer (approved):
LGTM. I think we want at least some minimum events for being able to construct history easily.
It probably helps to tools like Vulcanize. Btw `tend` and `dent` have the `LogNote` event.

PR #147, opened 2020-10-05
[draft] make bad debt grace period configurable per-Ilk

livnev:
`Vow.wait` exists in order to avoid creating unnecessary `flop` auctions immediately after a position has entered liquidation (creating bad debt), but before the `flip` auction has had time to finish and cancel the bad debt. Currently, `Vow.wait` has to be configured to apply to debt generated by all collateral types, meaning that it should be set to the maximum running time of any `flip` auction. This is a limitation, since some future collateral types may wish to allow much longer liquidation auction durations than the existing ones, which would force `wait` to be longer, unnecessarily delaying `flop` auctions for other collateral types.

This PR makes `wait` configurable per `Ilk`, as `Vow.ilks(ilk).wait`.

TODOs:
- adjust tests

### DIFF
--- a/src/cat.sol
+++ b/src/cat.sol
@@ -42,7 +42,7 @@ interface VatLike {
 }
 
 interface VowLike {
-    function fess(uint256) external;
+    function fess(uint256,uint256) external;
 }
 
 contract Cat is LibNote {
@@ -60,6 +60,7 @@ contract Cat is LibNote {
         address flip;  // Liquidator
         uint256 chop;  // Liquidation Penalty  [wad]
         uint256 dunk;  // Liquidation Quantity [rad]
+        uint256 wait;  // Bad debt grace period [seconds]
     }
 
     mapping (bytes32 => Ilk) public ilks;
@@ -116,6 +117,7 @@ contract Cat is LibNote {
     function file(bytes32 ilk, bytes32 what, uint256 data) external note auth {
         if (what == "chop") ilks[ilk].chop = data;
         else if (what == "dunk") ilks[ilk].dunk = data;
+        else if (what == "wait") ilks[ilk].wait = data;
         else revert("Cat/file-unrecognized-param");
     }
     function file(bytes32 ilk, bytes32 what, address flip) external note auth {
@@ -155,7 +157,7 @@ contract Cat is LibNote {
         vat.grab(
             ilk, urn, address(this), address(vow), -int256(dink), -int256(dart)
         );
-        vow.fess(mul(dart, rate));
+        vow.fess(milk.wait, mul(dart, rate));
 
         { // Avoid stack too deep
             // This calcuation will overflow if dart*rate exceeds ~10^14,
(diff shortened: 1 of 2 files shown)


gbalabasquer:
I know this is still a very early PR, but have 2 suggestions that might make sense:

- Keep the `wait` value per collateral in the `vow`. That way we might save replacing the `cat`. I haven't analysed if keeping the same `cat` is something convenient or which implications it might have, but at least the possibility of replacing just the `vow` is there (similar analysis should be done for `flapper` and `flopper`).

- Instead of saving the `tab` in the 2 dimensions `era` and `wait` mapping, just save it in one dimension mapping with the addition of both values.

livnev:
Hi @gbalabasquer, thanks for your attention and suggestions:

1. It seemed better to avoid introducing the concept of `Ilk`s (and per-`Ilk` data) in the `Vow`, seeing as currently the `Vow` does not concern itself with collateral-specific information. Also, even if we were to store per-`Ilk` `wait` in the `Vow`, we would still need a way of referencing which `Ilk` the `fess` is coming from (to know which `wait` to use), meaning that the interface would change and would require a change in the `Cat` anyway.

2. Good point, I have further simplified `fess` with this idea.

Overall, the main disadvantage of this change is the need to deploy a new `Vow` (and new `Cat`), which seems unavoidable. I wanted to flag this limitation, so that it can at least be incorporated the next time it is convenient to redeploy the `Vow` (if ever).

gbalabasquer:
1. Right, I forgot about having to identify the `ilk` wait time when executing `fess`.

2. Great.

Yes, as you know, replacing the `vow` is a bit complicate process. We made some analysis about it at the moment the `flop` stuck auctions happened. We ended up fixing that specific problem with some work around in the `flop`. However a more natural fix would be in the `vow`.
I guess at some point `vow` should be updated with all these improvements.

src/vow.sol
```diff
@@ -109,13 +108,13 @@ contract Vow is LibNote {
...
-    function fess(uint tab) external note auth {
-        sin[now] = add(sin[now], tab);
+    function fess(uint era, uint tab) external note auth {
+        sin[era] = add(sin[era], tab);
         Sin = add(Sin, tab);
     }
     // Pop from debt-queue
-    function flog(uint era) external note {
-        require(add(era, wait) <= now, "Vow/wait-not-finished");
+    function flog(uint wait, uint era) external note {
```
gbalabasquer, line 116: Missing removing `wait`

livnev: 👍

src/cat.sol
```diff
@@ -155,7 +157,7 @@ contract Cat is LibNote {
         vat.grab(
             ilk, urn, address(this), address(vow), -int256(dink), -int256(dart)
         );
-        vow.fess(mul(dart, rate));
+        vow.fess(add(now, milk.wait), mul(dart, rate));
```
godsflaw, line 160: Make sure these changes make it into the `dog.sol`

PR #148, opened 2020-10-05
feat: Add Auction Reset Tests (SC-6413)

lucas-manuel:
Add testing for `redo` function in `clip.sol`

### DIFF
--- a/src/test/clip.t.sol
+++ b/src/test/clip.t.sol
@@ -693,37 +693,30 @@ contract ClipperTest is DSTest {
         assertEq(vat.gem(ilk, me), 960 ether + lotReturn);  // Collateral returned (10 WAD)
     } 
 
-    function testFail_auction_reset_tail() public {
+    function auctionResetSetup(uint256 tau) internal {
         LinearDecrease calc = new LinearDecrease();
-        calc.file(bytes32("tau"), 36000);   // 10 hours till zero is reached (used to test tail)  
+        calc.file(bytes32("tau"), tau);   // 10 hours till zero is reached (used to test tail)  
 
         clip.file("buf",  ray(1.25 ether)); // 25% Initial price buffer
         clip.file("dust", rad(20   ether)); // $20 dust
         clip.file("calc", address(calc));   // File price contract
-        clip.file("cusp", ray(0.3 ether));  // 70% drop before reset
+        clip.file("cusp", ray(0.5 ether));  // 50% drop before reset
         clip.file("tail", 3600);            // 1 hour before reset
 
         assertEq(clip.kicks(), 0);
         dog.bark(ilk, me);
         assertEq(clip.kicks(), 1);
+    }
+
+    function testFail_auction_reset_tail() public {
+        auctionResetSetup(10 hours); // 10 hours till zero is reached (used to test tail)  
 
         hevm.warp(now + 3600 seconds);
         clip.redo(1);
     }
 
     function test_auction_reset_tail() public {
-        LinearDecrease calc = new LinearDecrease();
-        calc.file(bytes32("tau"), 36000);   // 10 hours till zero is reached (used to test tail)  
-
-        clip.file("buf",  ray(1.25 ether)); // 25% Initial price buffer
-        clip.file("dust", rad(20   ether)); // $20 dust
-        clip.file("calc", address(calc));   // File price contract
-        clip.file("cusp", ray(0.3 ether));  // 70% drop before reset
-        clip.file("tail", 3600);            // 1 hour before reset
-
-        assertEq(clip.kicks(), 0);
-        dog.bark(ilk, me);
-        assertEq(clip.kicks(), 1);
+        auctionResetSetup(10 hours); // 10 hours till zero is reached (used to test tail) 
 
         pip.poke(bytes32(uint256(3 ether))); // Spot = $1.50 (update price before reset is called)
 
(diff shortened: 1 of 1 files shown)


src/test/clip.t.sol
```diff
@@ -692,4 +692,85 @@ contract ClipperTest is DSTest {
         uint256 lotReturn = 30 ether - expectedGem;         // lot - loaf.tab / max = 15
         assertEq(vat.gem(ilk, me), 960 ether + lotReturn);  // Collateral returned (10 WAD)
     } 
+
+    function auctionResetSetup(uint256 tau) internal {
+        LinearDecrease calc = new LinearDecrease();
+        calc.file(bytes32("tau"), tau);   // 10 hours till zero is reached (used to test tail)
```
gbalabasquer, line 698:
```suggestion
        calc.file(bytes32("tau"), tau);   // tau hours till zero is reached (used to test tail)  
```

src/test/clip.t.sol
```diff
@@ -692,4 +692,85 @@ contract ClipperTest is DSTest {
...
+        hevm.warp(startTime + 3601 seconds);
+        clip.redo(1);
+        
+        (,,,, uint96 tic_after, uint256 top_after) = clip.sales(1);
+        assertEq(uint256(tic_after), startTime + 3601 seconds);     // (now)
+        assertEq(top_after, ray(3.75 ether)); // $3 spot + 25% buffer = $5 (used most recent OSM price)
+    }
+
+    function testFail_auction_reset_cusp() public {
+        auctionResetSetup(1 hours);    // 10 hours till zero is reached (used to test tail)
```
gbalabasquer, line 736:
```suggestion
        auctionResetSetup(1 hours);    // 1 hour till zero is reached (used to test tail) 
```

src/test/clip.t.sol
```diff
@@ -692,4 +692,85 @@ contract ClipperTest is DSTest {
...
+        clip.file("calc", address(calc));   // File price contract
+        clip.file("cusp", ray(0.5 ether));  // 50% drop before reset
+        clip.file("tail", 3600);            // 1 hour before reset
+
+        assertEq(clip.kicks(), 0);
+        dog.bark(ilk, me);
+        assertEq(clip.kicks(), 1);
+    }
+
+    function testFail_auction_reset_tail() public {
```
gbalabasquer, line 711: This test is correct, however it doesn't check exactly the same than the next one.
I mean I could expect the `pip.poke` had something to do with the next test passing, and not the extra second in the `hevm.warp`.
In general the `testFail` is not very informative. We might want to start checking failures everywhere wrapping everything in `try_call` kind functions and incorporating them in the same test, like:
```
hevm.warp(startTime + 3600 seconds);
assertEq(!try_redo(1));
hevm.warp(startTime + 3601 seconds);
assertEq(try_redo(1));
```

Anyway if we want to convert all the tests to that way we could do a new PR.

gbalabasquer:
I think I have some suggestions to standardize a bit these helper functions that do repetitive things.
But will do that in another PR when this one gets merged.

src/test/clip.t.sol
```diff
@@ -692,4 +692,85 @@ contract ClipperTest is DSTest {
...
+        auctionResetSetup(10 hours); // 10 hours till zero is reached (used to test tail) 
+        
+        hevm.warp(startTime + 3601 seconds);
+        clip.redo(1);
+
+        clip.redo(1);
+    }
+
+    function testFail_auction_reset_cusp_twice() public {
+        auctionResetSetup(1 hours); // 1 hours till zero is reached (used to test cusp)
```
gbalabasquer, line 769:
```suggestion
        auctionResetSetup(1 hours); // 1 hour till zero is reached (used to test cusp) 
```

src/test/clip.t.sol
lucas-manuel: Fixed

src/test/clip.t.sol
lucas-manuel: Fixed

src/test/clip.t.sol
```diff
@@ -692,4 +692,85 @@ contract ClipperTest is DSTest {
...
+
+    function testFail_auction_reset_cusp() public {
+        auctionResetSetup(1 hours);    // 10 hours till zero is reached (used to test tail) 
+
+        hevm.warp(now + 1800 seconds); // 50% drop after 30min, price is at "cusp"
+        clip.redo(1);
+    }
+
+    function test_auction_reset_cusp() public {
+        auctionResetSetup(1 hours); // 1 hours till zero is reached (used to test cusp)
```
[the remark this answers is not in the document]
lucas-manuel: Fixed

src/test/clip.t.sol
lucas-manuel: Fixed

lucas-manuel:
Added @gbalabasquer's suggestions.

gbalabasquer (approved):
LGTM

src/test/clip.t.sol
```diff
@@ -692,4 +692,80 @@ contract ClipperTest is DSTest {
...
+        string memory sig = "redo(uint256)";
+        (ok,) = address(clip).call(abi.encodeWithSignature(sig, id));
+    }
+
+    function test_auction_reset_tail() public {
+        auctionResetSetup(10 hours); // 10 hours till zero is reached (used to test tail) 
+
+        pip.poke(bytes32(uint256(3 ether))); // Spot = $1.50 (update price before reset is called)
+
+        (,,,, uint96 tic_before, uint256 top_before) = clip.sales(1);
```
kmbarry1, line 721: I hate to be that guy...but we generally `camelCase` variable names (see e.g. `startTime`, and other values in this file), and I'd like to try to maintain that style, or at the very least have each source file be self-consistent in its formatting conventions. Mind changing `tic_before` ==> `ticBefore`, etc?

src/test/clip.t.sol
```diff
@@ -692,4 +692,80 @@ contract ClipperTest is DSTest {
...
+        assertEq(top_before, ray(5 ether)); // $4 spot + 25% buffer = $5 (wasn't affected by poke)
+        
+        hevm.warp(startTime + 1800 seconds);
+        assertTrue(!try_redo(1));
+        hevm.warp(startTime + 1801 seconds);
+        assertTrue( try_redo(1));
+        
+        (,,,, uint96 tic_after, uint256 top_after) = clip.sales(1);
+        assertEq(uint256(tic_after), startTime + 1801 seconds);     // (now)
+        assertEq(top_after, ray(3.75 ether)); // $3 spot + 25% buffer = $5 (used most recent OSM price)
```
kmbarry1, line 751: Should `$5` be `$3.75` here?

src/test/clip.t.sol
kmbarry1: This is a good point...every time I write a `testFail`, I have to comment out the line that's supposed to revert and make sure it fails without that line...and some change that makes an earlier line revert can hide a regression in the behavior of the later call. It's just a fragile pattern.

I think we just discovered some technical debt :) (since we use `testFail` in more places than here).

src/test/clip.t.sol
lucas-manuel: Yeah I can do that 👍

lucas-manuel:
Addressed @kmbarry1's comments

kmbarry1 (approved):
LGTM

PR #150, opened 2020-10-09
SC-4636

andy8052:
This is an updated circuit breaker logic. 

The code has 4 levels of `breaker`:
0: no breaker
1: no new kick()
2: no new warm() or take()
3: yank() can be called

Any new logic can pass a breaker level (a uint256) and a boolean (higher or lower than a level) to `isStopped` (this probably needs a new name but I really suck at function naming. This allows for a very modular and governance controlled breaker system for the auctions. Maybe it is too open ended?

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -83,6 +83,12 @@ contract Clipper {
 
     uint256 internal locked;
 
+    // Levels for circuit breaker
+    // 0: no breaker
+    // 1: no new kick()
+    // 2: no new warm() or take()
+    uint256 public stopped = 0;
+
     // --- Events ---
     event Kick(
         uint256  id,
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -130,6 +130,11 @@ contract Clipper {
         locked = 0;
     }
 
+    modifier isStopped(uint256 level, bool which) {
+        require(stopped < level == which, "Clipper/stopped-incorrect");
+        _;
+    }
+
     // --- Administration ---
     function file(bytes32 what, uint256 data) external {
         if      (what ==  "buf") buf  = data;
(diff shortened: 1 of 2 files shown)


src/test/clip.t.sol
```diff
@@ -801,4 +801,80 @@ contract ClipperTest is DSTest {
...
+        hevm.warp(startTime + 3600 seconds);
+        assertTrue(!try_redo(1));
+        hevm.warp(startTime + 3601 seconds);
+        assertTrue( try_redo(1));
+    }
+
+    // does nothing but we can still test the breaker works
+    function test_yank() public {
+        clip.setBreaker(3);
+        clip.yank();
```
kmbarry1, line 873: As I mentioned on your other PR, `yank` shouldn't require the circuit breaker to be turned on. Or, at least there should be a version of it that is just `auth` for the `End` to call. You can, for example, make a wrapper around it and have that require the breaker to be on (so long as you also `rely` the contract on itself). Or just have two functions.

andy8052: Ah yeah my bad forgot about this

lucas-manuel (approved):
LGTM

src/clip.sol
```diff
@@ -124,6 +130,11 @@ contract Clipper {
         locked = 0;
     }
 
+    modifier isStopped(uint256 level, bool which) {
```
kmbarry1, line 133: I really don't think the second arg is necessary now that we have the levels...we're just passing it as true everywhere anyway.

andy8052: Yeah so with yank it was false. If we ever wanted a yank where we were say it has to be 3 or more then we want it. I was unsure on whether or not I wanted to keep it

kmbarry1: I would keep things simple--there's no need for it right now, it makes the code harder to read/less gas efficient, and we can add it back if we have to. `yank` could always implement some more custom condition without a modifier as a special case.

andy8052: Sounds good will pull it

kmbarry1 (approved):
LGTM

PR #152, opened 2020-10-15
minor gas optimizations in Clipper

kmbarry1:
Should be an overall optimization (not exactly sure how `price * slice == tab` case is affected w/o profiling, but that should be the rarest case by far, and all other cases are strictly improved).

Maybe not worth the trouble as storage operations overwhelmingly dominate gas costs anyway, but thought it was worth taking a look at.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -255,6 +255,7 @@ contract Clipper {
         } else {  // owe < sale.tab
             // Calculate remaining tab after operation
             sale.tab = sale.tab - owe;  // safe since owe < sale.tab
+            (,,,, uint256 dust) = vat.ilks(ilk);
             require(sale.tab >= dust, "Clipper/dust");
         }
 
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -242,26 +242,25 @@ contract Clipper {
         require(max >= price, "Clipper/too-expensive");
 
         // Purchase as much as possible, up to amt
-        uint256 slice = min(sale.lot, amt);
+        uint256 slice = min(sale.lot, amt);  // slice <= lot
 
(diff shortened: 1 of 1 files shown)


kmbarry1:
With this PR:
```
[PASS] test_take_over_tab() (gas: 1168951)
[PASS] test_take_at_tab() (gas: 1168884)
[PASS] test_take_under_tab() (gas: 1144056)
[PASS] test_take_multiple_bids_different_prices() (gas: 1306054)
```

Without this PR:
```
[PASS] test_take_over_tab() (gas: 1169155)
[PASS] test_take_at_tab() (gas: 1169038)
[PASS] test_take_under_tab() (gas: 1144243)
[PASS] test_take_multiple_bids_different_prices() (gas: 1306445)
```

Deltas:
`test_take_over_tab` : -204
`test_take_at_tab` : -154
`test_take_under_tab` : -187
`test_take_multiple_bids_different_prices` : -391

So seems to be a strict improvement, albeit small. For comparison, an `SLOAD` costs 200 gas.

lucas-manuel (dismissed):

kmbarry1:
note: should see if a test can be added to catch any mutation of the code that makes the lack of safe math unsafe

kmbarry1:
So, it's not really feasible to write a unit test that catches an arbitrary code change that breaks the no-underflow invariant, basically because there's no reasonably-sized set of inputs that would catch all possible mutations. Fortunately, this is exactly why we formally verify contracts--something breaking that invariant will always break a properly-written formal spec.

After all the back-and-forth, I'm in favor of merging.

lucas-manuel:
I agree, let's merge.

lucas-manuel (approved):

PR #153, opened 2020-10-19
feat: Add Keeper Liquidation Incentive (SC-6597)

lucas-manuel:
Making this PR to start discussing how to incentivize keepers to liquidate vaults in LIQ-2.0.

I have considered the following approaches:

1. Use `vat.suck()` in `dog.bark()` with a governance parameter `tip` set as a constant value of `dai` used for all ilks.
2. Use `vat.suck()` in `dog.bark()` with a governance parameter `tip` set as a constant value of `dai` used per ilk.
3. Use `vat.suck()` in `dog.bark()` with a governance parameter `tip` set as a percentage of `due` for a `barked` vault used for all ilks.
4. Use `vat.suck()` in `dog.bark()` with a governance parameter `tip` set as a percentage value of `due` for a `barked` vault used per ilk.
5. Dedicate a portion of `chop` for liquidators to claim on successful liquidation.
6. Flux a portion of collateral to liquidator on `dog.bark()`.
7. Gas refund token functionality.

I have implemented option 1 in this PR as I believe it is the simplest solution and incentivizes keepers in a fair manner, but I am very open to other suggestions and general discussion about the optimal solution.

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -60,9 +60,11 @@ contract Dog is LibNote {
     // --- Data ---
     struct Ilk {
         address clip;  // Liquidator
-        uint256 chop;  // Liquidation Penalty  [wad]
+        uint256 chop;  // Liquidation Penalty                                          [wad]
         uint256 hole;  // Max DAI needed to cover debt+fees of active auctions per ilk [rad]
         uint256 dirt;  // Amt DAI needed to cover debt+fees of active auctions per ilk [rad]
+        uint256 chip;  // Percentage of tab to suck from vat to incentivize keepers    [wad]
+        uint256 tip;   // Flat fee to suck from vat to incentivize keepers             [rad]
     }
 
     mapping (bytes32 => Ilk) public ilks;
@@ -70,7 +72,6 @@ contract Dog is LibNote {
     uint256 public live;  // Active Flag
     VatLike public vat;   // CDP Engine
     VowLike public vow;   // Debt Engine
-    uint256 public tip;   // Keeper liquidation incentive                         [rad]
     uint256 public Hole;  // Max DAI needed to cover debt+fees of active auctions [rad]
     uint256 public Dirt;  // Amt DAI needed to cover debt+fees of active auctions [rad]
 
@@ -107,6 +108,9 @@ contract Dog is LibNote {
     function mul(uint256 x, uint256 y) internal pure returns (uint256 z) {
         require(y == 0 || (z = x * y) / y == x);
     }
+    function wmul(uint x, uint y) internal pure returns (uint z) {
+        z = add(mul(x, y), WAD / 2) / WAD;
+    }
 
     // --- Administration ---
     function file(bytes32 what, address data) external note auth {
@@ -115,12 +119,13 @@ contract Dog is LibNote {
     }
     function file(bytes32 what, uint256 data) external note auth {
         if (what == "Hole") Hole = data;
-        else if (what == "tip") tip = data;
         else revert("Dog/file-unrecognized-param");
     }
     function file(bytes32 ilk, bytes32 what, uint256 data) external note auth {
         if (what == "chop") ilks[ilk].chop = data;
         else if (what == "hole") ilks[ilk].hole = data;
+        else if (what == "chip") ilks[ilk].chip = data;
+        else if (what == "tip")  ilks[ilk].tip  = data;
         else revert("Dog/file-unrecognized-param");
     }
     function file(bytes32 ilk, bytes32 what, address clip) external note auth {
@@ -181,9 +186,9 @@ contract Dog is LibNote {
                 lot: dink,
                 usr: urn
             });
+            
+            vat.suck(address(vow), msg.sender, add(milk.tip, wmul(tab, milk.chip)));
         }
-
-        vat.suck(address(vow), msg.sender, tip);
 
         emit Bark(ilk, urn, dink, dart, due, milk.clip, id);
     }
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -109,7 +109,7 @@ contract Dog is LibNote {
         require(y == 0 || (z = x * y) / y == x);
     }
     function wmul(uint x, uint y) internal pure returns (uint z) {
-        z = add(mul(x, y), WAD / 2) / WAD;
+        z = mul(x, y) / WAD;
     }
 
     // --- Administration ---
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -38,6 +38,7 @@ interface VatLike {
         uint256 art   // [wad]
     );
     function grab(bytes32,address,address,address,int256,int256) external;
+    function suck(address,address,uint256) external;
     function hope(address) external;
     function nope(address) external;
 }
@@ -69,6 +70,7 @@ contract Dog is LibNote {
     uint256 public live;  // Active Flag
     VatLike public vat;   // CDP Engine
     VowLike public vow;   // Debt Engine
+    uint256 public tip;   // Keeper liquidation incentive                         [rad]
     uint256 public Hole;  // Max DAI needed to cover debt+fees of active auctions [rad]
     uint256 public Dirt;  // Amt DAI needed to cover debt+fees of active auctions [rad]
 
@@ -113,6 +115,7 @@ contract Dog is LibNote {
     }
     function file(bytes32 what, uint256 data) external note auth {
         if (what == "Hole") Hole = data;
+        else if (what == "tip") tip = data;
         else revert("Dog/file-unrecognized-param");
     }
     function file(bytes32 ilk, bytes32 what, uint256 data) external note auth {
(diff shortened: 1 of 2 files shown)


kmbarry1:
The nice thing about a constant reward is that it's equally profitable to liquidate small versus large Vaults, so from a Vault holder perspective it's fair.

We should think carefully about the potential for abuse (e.g. opening Vaults and liquidating them to farm the liquidation incentive), but at first blush it seems safe since 1) there's no way to guarantee you'll win the race to call `bark()` and 2) having both `dust` and `chop` set well makes this unprofitable (basically the `chop` on a borderline-dusty Vault needs to exceed `tip`). But e.g. collaterals with e.g. very low collateralization ratios might be vulnerable to this (looking at you, `USDC-A`).

lucas-manuel:
I have another possible approach that we could take:

We could have two governance params, one to calculate a portion of `due` as an incentive (`cut` in this example), and as a flat incentive (`fee` in this example). Every time we liquidate a vault we could pay the liquidator this amount: `min(rmul(due, cut), fee)`

Where `cut` could be a number like 0.98 RAY and `fee` would be a flat DAI amount in RAD. This would make it explicitly unprofitable for a keeper to liquidate their own dusty vault maliciously, but would allow us to set a higher flat incentive for keepers to liquidate larger CDPs.

`cut` and `fee` could both either be set per ilk or across all ilks, so we should discuss which would be ideal for both if we choose this approach.

kmbarry1:
Thoughts on the `cut`/`fee` idea:

I'm not necessarily sure we want to do what amounts to disincentivizing liquidations of very small Vaults--that might perversely encourage people to open a bunch of Vaults just barely above whatever the current dust limit is since those Vaults will be less attractive to liquidators, and thus they might be able to escape liquidation. Although I think at the end of the day so long as `dust` and other parameters are set well, it's safe.

Btw, wanted to offer a counterpoint to the fixed-reward model: from the system's perspective (and thus from the DAI _and_ MKR holder perspectives), a large undercollateralized Vault is riskier than a small undercollateralized Vault, so maybe the reward for big liquidations _should_ be larger than for small liquidations. Further, the extra system income from `chop` is also proportional to the size of the the Vault, so in theory the system can _afford_ to pay more for larger liquidations as well.

The natural generalization of the two models is of course some linear combination of a scaling reward and fixed reward, so if there's no clear consensus on which is better, we could use that (at the expense of higher gas for an extra storage read + math operations). Which is the direction @lucas-manuel is going with the `cut`/`fee` suggestion anyway.

lucas-manuel:
Building on @kmbarry1's comment, we could actually maybe make the bottom end the constant. So the minimum amount would be something like `rmul(rmul(dust, sub(mat, RAY)), 0.98 RAY)`, so for example in an ETH-A vault if someone adds 150 as collateral, takes out 100 DAI and liquidates themselves, they would net 149 DAI, which would be a loss. But the other value could be a percentage of chop that will scale indefinitely with the size of the vault.

kmbarry1:
I'm beginning to think it might be better, whatever we do, to have some sort of per-ilk component, simply because `chop` and `mat` are so relevant to tuning the liquidation reward (and collateral price might be as well), and those are per-collateral. Seems like we're excessively complicating it by trying to define a global formula--no matter how clever we make it, there will probably eventually be some future market conditions for which it is suboptimal for at least one collateral type.

So, maybe I'd propose, per-collateral, two parameters:
`tip` : constant amount awarded per-liquidation
`cut` : multiplicative factor dependent on amount of debt (essentially a portion of `chop`)

In most cases, `tip` would be zero and just `cut` would be used. In rare cases in which a specific incentive, say, to liquidate small Vaults was needed, `tip` could be set to an appropriate amount, combined e.g. with temporarily disabling new Vault creation for that ilk to prevent farming the liquidation reward (although it's an unsolved problem how to get people to bid in the resulting auctions; MKR holders could probably arrange out-of-band to pay Keepers for that service).

lucas-manuel:
I agree that having the parameters set per-ilk would be better. Are you suggesting something like `reward = tip + cut * chop`?

kmbarry1:
More like:

`reward = tip + wmul(tab, cut)`

I guess you could also do:

`reward = tip +wmul(tab, wmul(chop, cut))`

to try to enforce that the linear component is always less than `chop`, but governance could just set `cut > WAD`...you could add validation in `file` to prevent that, but generally the design philosophy in `dss` has been that, at a certain point, you have to trust governance to set correct parameters, so checking things to death just isn't worth the time, because there are lots of places where you have to leave sufficient flexibility for market conditions/special circumstances anyway. That philosophy leads to e.g. the extensive testing + tooling for spells that we're developing, and that's better since because it's off-chain it can be as detailed and mutable as necessary.

lucas-manuel:
> More like:
> 
> `reward = tip + wmul(tab, cut)`
> 
> I guess you could also do:
> 
> `reward = tip +wmul(tab, wmul(chop, cut))`
> 
> to try to enforce that the linear component is always less than `chop`, but governance could just set `cut > WAD`...you could add validation in `file` to prevent that, but generally the design philosophy in `dss` has been that, at a certain point, you have to trust governance to set correct parameters, so checking things to death just isn't worth the time, because there are lots of places where you have to leave sufficient flexibility for market conditions/special circumstances anyway. That philosophy leads to e.g. the extensive testing + tooling for spells that we're developing, and that's better since because it's off-chain it can be as detailed and mutable as necessary.

Yes sorry I was trying to express that second equation (didn't have enough coffee 😆 ). I am actually more of a fan of the first equation now that I think of it, because there could be an edge case where governance wants to incentivize liquidations so bad that they are willing to do it at a loss to the protocol. I know that this can also be accomplished by the second equation by setting `cut > WAD`, but if we can avoid the extra `wmul` we might as well. Also if we are filing `cut` on a per ilk basis it wouldn't be hard to calculated using the `chop` and test against it.

Also just a note we would want to make that second equation:
`reward = tip + wmul(tab, wmul(sub(chop, RAY), cut))` if we used it.

lucas-manuel:
Changed incentive to `vat.suck(address(vow), msg.sender, add(milk.tip, wmul(tab, milk.chip)));` in the last commit.

Using `chip` since we use `cut` in the exponential decrease in `abaci.sol`

src/dog.sol
```diff
@@ -105,6 +108,9 @@ contract Dog is LibNote {
     function mul(uint256 x, uint256 y) internal pure returns (uint256 z) {
         require(y == 0 || (z = x * y) / y == x);
     }
+    function wmul(uint x, uint y) internal pure returns (uint z) {
+        z = add(mul(x, y), WAD / 2) / WAD;
```
godsflaw, line 112: Let's replace this with a simple `wmul()`:
```
function rmul(uint x, uint y) internal pure returns (uint z) {
        z = mul(x, y) / WAD;
}
```

godsflaw (approved):
LGTM with an added pre-check for Bob's balance in the test.

brianmcmichael (approved):

src/dog.sol
```diff
@@ -59,9 +60,11 @@ contract Dog is LibNote {
     // --- Data ---
     struct Ilk {
         address clip;  // Liquidator
-        uint256 chop;  // Liquidation Penalty  [wad]
+        uint256 chop;  // Liquidation Penalty                                          [wad]
         uint256 hole;  // Max DAI needed to cover debt+fees of active auctions per ilk [rad]
         uint256 dirt;  // Amt DAI needed to cover debt+fees of active auctions per ilk [rad]
+        uint256 chip;  // Percentage of tab to suck from vat to incentivize keepers    [wad]
+        uint256 tip;   // Flat fee to suck from vat to incentivize keepers             [rad]
```
kmbarry1, line 67: I'd personally phrase the comments as "suck from vow" as that's where the `sin` is assigned.

kmbarry1 (approved):
LGTM

After merging, I'll add these to the public forum post of risk parameters (and probably update the technical summary too) and advertise to the community so we can get feedback.

We'll probably do a risk parameter call w/the community at some point as well.

lucas-manuel:
@kmbarry1 Updated comments as requested.

src/dog.sol
```diff
@@ -178,6 +186,8 @@ contract Dog is LibNote {
                 lot: dink,
                 usr: urn
             });
+            
+            vat.suck(address(vow), msg.sender, add(milk.tip, wmul(tab, milk.chip)));
```
kmbarry1, line 190: Just realized this should use `due`, not `tab`. Otherwise it's very confusing to set `chip` correctly, as you have to do an extra correction, potentially, for taking part of `chop` as well.

lucas-manuel: Good point. Do you think I can push this change straight to the `liq-2.0` branch or should I make a PR?

kmbarry1: Mhm, you'll probably need to tweak the tests too (at least, I hope so, otherwise we need a test to cover this)...better to do a PR then IMO.

lucas-manuel: Yeah good call will do

PR #154, opened 2020-10-20
feat: Add Dusty `bite` Check (SC-6351)

lucas-manuel:
Adds a check to ensure that when `dog.bark()` is called,  the liquidated CDP is not left in a dusty state.

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -92,6 +92,7 @@ contract Dog /* is LibNote */ {
 
     // --- Math ---
     uint256 constant WAD = 10 ** 18;
+    uint256 constant RAY = 10 ** 27;
 
     function min(uint256 x, uint256 y) internal pure returns (uint256 z) {
         if (x > y) { z = y; } else { z = x; }
@@ -148,6 +149,9 @@ contract Dog /* is LibNote */ {
             require(room > 0 && room >= dust, "Dog/liquidation-limit-hit");
 
             dart = min(art, mul(room, WAD) / rate / milk.chop);
+
+            // Verify that CDP is not left in a dusty state
+            require(dart == art || mul(art - dart, RAY) >= dust, "Dog/leaves-dust");
         }
 
         uint256 dink = min(ink, mul(ink, dart) / art);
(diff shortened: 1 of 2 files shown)


src/dog.sol
```diff
@@ -148,14 +149,16 @@ contract Dog /* is LibNote */ {
             require(room > 0 && room >= dust, "Dog/liquidation-limit-hit");
 
             dart = min(art, mul(room, WAD) / rate / milk.chop);
+
+            // Verify that CDP is not left in a dusty state
+            require(dart == art || mul(art - dart, RAY) >= dust, "Dog/leaves-dust");
```
kmbarry1, line 154: Need to multiply by `rate` here, not `RAY`, because `dust` is measured in dai, not in normalized debt.

kmbarry1 (changes requested):
I think just reverting here is sufficient, as in 2.0 this only happens due to `hole` or `Hole` being nearly at its limit, which is a rare scenario anyway.

kmbarry1 (changes requested):
Consider adding a test to cover the `rate =/= RAY` case that was fixed in the most recent commit.

src/dog.sol
```diff
@@ -92,6 +92,7 @@ contract Dog /* is LibNote */ {
 
     // --- Math ---
     uint256 constant WAD = 10 ** 18;
+    uint256 constant RAY = 10 ** 27;
```
kmbarry1, line 95:
```suggestion
```

src/test/clip.t.sol
```diff
@@ -438,6 +446,109 @@ contract ClipperTest is DSTest {
...
+        uint256 lot;
+        address usr;
+        uint96  tic;
+        uint256 top;
+        uint256 ink;
+        uint256 art;
+
+        jug.file(ilk, "duty", 1000000001243680656318820312);
+        hevm.warp(now + 1 days);
+        jug.drip(ilk);
```
kmbarry1, line 512: Please just call `Vat.fold` directly to set `rate`, it keeps the tests simpler by avoiding the `Jug` dependency.

kmbarry1 (approved):
LGTM

PR #156, opened 2020-10-23
feat: Add LibNote events to dog.sol, clip.sol (SC-4645)

lucas-manuel:
Adding `note` modifiers to all relevant functions in LIQ-2.0. This PR will also serve as a place to discuss parameters used for custom events in LIQ-2.0 functions, and if any other data needs to be captured for VulcanizeDB.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -92,21 +92,26 @@ contract Clipper is LibNote {
 
     // --- Events ---
     event Kick(
-        uint256  id,
+        uint256 indexed id,
+        uint256 top,
         uint256 tab,
         uint256 lot,
         address indexed usr
     );
 
     event Take(
-        uint256  id,
+        uint256 indexed id, 
+        uint256 max,
+        uint256 price,
+        uint256 owe,
         uint256 tab,
         uint256 lot,
         address indexed usr
     );
 
     event Redo(
-        uint256  id,
+        uint256 indexed id,
+        uint256 top,
         uint256 tab,
         uint256 lot,
         address indexed usr
@@ -196,7 +201,7 @@ contract Clipper is LibNote {
         require(has, "Clipper/invalid-price");
         sales[id].top = rmul(rdiv(mul(uint256(val), BLN), spot.par()), buf);
 
-        emit Kick(id, tab, lot, usr);
+        emit Kick(id, sales[id].top, tab, lot, usr);
     }
 
     // Reset an auction
@@ -220,7 +225,7 @@ contract Clipper is LibNote {
         require(has, "Clipper/invalid-price");
         sales[id].top = rmul(rdiv(mul(uint256(val), 10 ** 9), spot.par()), buf);
 
-        emit Redo(id, sales[id].tab, sales[id].lot, sales[id].usr);
+        emit Redo(id, sales[id].top, sales[id].tab, sales[id].lot, sales[id].usr);
     }
 
     // Buy amt of collateral from auction indexed by id
@@ -288,7 +293,7 @@ contract Clipper is LibNote {
             sales[id].lot = sale.lot;
         }
 
-        emit Take(id, sale.tab, sale.lot, sale.usr);
+        emit Take(id, max, price, owe, sale.tab, sale.lot, sale.usr);
     }
 
     function _remove(uint256 id) internal {
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -17,6 +17,8 @@
 
 pragma solidity >=0.5.12;
 
+import "./lib.sol";
+
 interface VatLike {
     function move(address,address,uint256) external;
     function flux(bytes32,address,address,uint256) external;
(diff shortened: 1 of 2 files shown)


src/clip.sol
```diff
@@ -314,12 +316,12 @@ contract Clipper {
     }
 
     // --- Shutdown ---
-    function setBreaker(uint256 level) external auth {
+    function setBreaker(uint256 level) external note auth {
```
kmbarry1, line 319: It's worth considering whether triggering the circuit breaker merits its own custom event--in principle `note` is enough, but IIRC custom events are usually a little friendlier to integrators, and the breaker getting tripped would be a pretty important event to be able to reliably detect.

src/clip.sol
```diff
@@ -227,7 +229,7 @@ contract Clipper {
                   uint256 max,          // Maximum acceptable price (DAI / collateral) [ray]
                   address who,          // Receiver of collateral, payer of DAI, and external call address
                   bytes calldata data   // Data to pass in external call; if length 0, no call is done
-    ) external lock isStopped(2) {
+    ) external note lock isStopped(2) {
```
KentonPrescott, line 232: Comment regarding the custom Take event `emit Take(id, sale.tab, sale.lot, sale.usr);`

I recommend adding `owe` and `price` to improve the integrator's experience of determining participation size and price, respectively. Although one could reconstruct those values, with knowledge of previous tab / lots and reading the price curve, as defined in the `abacus`, I think it adds unnecessary friction for the shallow integration / analysis project. 

IIUC, `LogNote` does not record more than two function parameters, so I also think that `max` should be added, as it would be an interesting data point to consider when inspecting keeper behavior.

src/clip.sol
KentonPrescott: Custom events are friendlier in a couple of ways:
1. Can query the log directly rather than needing to query all `LogNotes` and filtering all the with the desired `msg.sig`
2. Has the privilege of logging information that was computed within the function

To my surprise, however, it looks like pulling historical logs or listening to an active event is easier than I expected. [Here's an example in web3.py of setting up an event filter](https://web3py.readthedocs.io/en/stable/filters.html#event-log-filters), which applies a second filter in the `topics` field. If it's as easy as it looks, then my first point is invalid..

KentonPrescott (commented):
Dropped a few comments, mostly around the custom events for some of the main functions. Regarding `clip.sol`, I think it would be nice to add the `indexed` keyword to the `id` parameter of all events, so it's easier to filter for a particular auction.

src/clip.sol
```diff
@@ -198,7 +200,7 @@ contract Clipper {
     }
 
     // Reset an auction
-    function redo(uint256 id) external isStopped(2) {
+    function redo(uint256 id) external note isStopped(2) {
```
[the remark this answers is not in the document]
lucas-manuel: I'm going to add `top` to the `Kick` event as well for the same reason.

src/clip.sol
lucas-manuel: Adding `owe`, `price` and `max`.

lucas-manuel:
Made changes requested by @KentonPrescott, still open to adding a custom circuit breaker event if need be.

src/clip.sol
lucas-manuel: Should we make a custom event anyways? Or would `note` be enough for integrators?

kmbarry1: I'm leaning towards `note` being good enough...for comparison, we don't have custom events for `rely`/`deny`/`file`, which are all fairly sensitive and important to monitor. `setBreaker` also lacks any internal values that we might otherwise want to expose.

kmbarry1 (approved):
LGTM

kmbarry1:
IMO this is ready to merge, unless @KentonPrescott has more feedback. We're still in a place where we can make tweaks, anyway.

src/clip.sol
KentonPrescott: +1 on the `note` being good enough for `setBreaker`

lucas-manuel:
> +1 on the `note` being good enough for `setBreaker` and the admin functions

@KentonPrescott replied above so we should be okay to merge if you want @kmbarry1

KentonPrescott:
It may be a good idea to take out `note` in `kick`, `redo`, and `take`.

Because the custom `Kick` and `Redo` events have everything a `LogNote` would offer, `note` may not be needed anymore. The custom `Take` event could be tweaked to reach the same result with the addition of `amt` and `who`. I've left out `data`, as I don't think it would be useful.

kmbarry1:
Given the importance of gas it seems wise to not do double logging. The question then becomes whether we should abandon `note` completely and just use custom events, or leave it mixed (which makes the code shorter/simpler but less consistent, and is maybe slightly more confusing for integrators).

cc @gbalabasquer

kmbarry1:
I'm fine with the mixture of `note` and custom events here, for the record. E.g. `Flip.kick` lacks a `note` modifier but emits a custom event, so a precedent exists.

KentonPrescott:
Just my 2 cents: I think mixed logging is better at this time because 1) it makes the code shorter/simpler and 2) is consistent with the some of `dss` (e.g. `Spotter` and `Dai`)

lucas-manuel:
Just removed `note` from `kick()`, `redo()`, `take()` and `bark()`, so now we are using mixed logging in the same way as `LIQ-1.2`.

kmbarry1 (approved):
LGTM

gbalabasquer:
I honestly prefer not mixing logs and removing `LogNote` at all. We can keep discussing it and in worst case make a new PR.

kmbarry1:
> I honestly prefer not mixing logs and removing `LogNote` at all. We can keep discussing it and if necessary, make a new PR. Take into account `DssGov` e.g. uses custom events for `rely/deny` and `file`.

Yeah, always possible to make another PR :). I think for this one we at least made a solid improvement--custom events where they are _needed_, with all the useful information in them, and _some_ logging everywhere. And like I said, we already have a precedent of mixing `note` + custom events in `dss`, and no one is really complaining about it, so maybe it's fine.

PR #157, opened 2020-10-30
feat: Use custom events

lucas-manuel:


src/clip.sol
```diff
@@ -91,14 +89,19 @@ contract Clipper is LibNote {
     uint256 public stopped = 0;
 
     // --- Events ---
+    event Rely(address indexed usr);
+    event Deny(address indexed usr);
+
+    event File(bytes32 indexed what, uint256 data);
+    event File(bytes32 indexed what, address data);
```
kmbarry1, line 96: I might actually name these differently to prevent confusion, e.g. `FileUint` and `FileAddress`.

src/dog.sol
```diff
@@ -73,6 +71,14 @@ contract Dog is LibNote {
     uint256 public Dirt;  // Amt DAI needed to cover debt+fees of active auctions [rad]
 
     // --- Events ---
+    event Rely(address indexed usr);
+    event Deny(address indexed usr);
+
+    event File(bytes32 indexed what, uint256 data);
+    event File(bytes32 indexed what, address data);
+    event File(bytes32 indexed ilk, bytes32 indexed what, uint256 data);
+    event File(bytes32 indexed ilk, bytes32 indexed what, address clip);
```
kmbarry1, line 80: I'd differentiate these as well with naming.

src/dog.sol
```diff
@@ -107,22 +115,26 @@ contract Dog is LibNote {
...
         if (what == "Hole") Hole = data;
         else revert("Dog/file-unrecognized-param");
+        emit File(what, data);
     }
-    function file(bytes32 ilk, bytes32 what, uint256 data) external note auth {
+    function file(bytes32 ilk, bytes32 what, uint256 data) external auth {
         if (what == "chop") ilks[ilk].chop = data;
         else if (what == "hole") ilks[ilk].hole = data;
         else revert("Dog/file-unrecognized-param");
+        emit File(what, data);
```
kmbarry1, line 132: Needs an `ilk` argument?

src/dog.sol
```diff
@@ -107,22 +115,26 @@ contract Dog is LibNote {
...
     }
-    function file(bytes32 ilk, bytes32 what, address clip) external note auth {
+    function file(bytes32 ilk, bytes32 what, address clip) external auth {
         if (what == "clip") ilks[ilk].clip = clip;
         else revert("Dog/file-unrecognized-param");
+        emit File(what, clip);
```
kmbarry1, line 137: Needs an `ilk` argument?

kmbarry1 (approved):
LGTM

src/clip.sol
```diff
@@ -117,6 +119,9 @@ contract Clipper is LibNote {
         address indexed usr
     );
 
+    event SetBreaker(uint256 level);
+    event Yank();
+
     // --- Init ---
     constructor(address vat_, address spot_, address dog_, bytes32 ilk_) public {
         vat  = VatLike(vat_);
```
gbalabasquer, line 127: Please add `emit Rely(msg.sender);` in the constructor.

src/clip.sol
```diff
@@ -91,14 +89,19 @@ contract Clipper is LibNote {
     uint256 public stopped = 0;
 
     // --- Events ---
+    event Rely(address indexed usr);
+    event Deny(address indexed usr);
+
+    event FileUint256(bytes32 indexed what, uint256 data);
```
gbalabasquer, line 95: Why not calling both just `File`? The signature of the event is the name + the parameters so shouldn't be any problem to reuse the name.

src/clip.sol
gbalabasquer: Ahh sorry just saw this. I actually prefer both with `File` name, but not opposed if you guys prefer to differentiate them.

src/clip.sol
gbalabasquer: Just saw the comments from Kurt. Left my comment there.

src/dog.sol
```diff
@@ -82,6 +88,8 @@ contract Dog is LibNote {
       address clip,
       uint256 id
     );
+    event Digs(bytes32 indexed ilk, uint256 rad);
+    event Cage();
 
     // --- Init ---
     constructor(address vat_) public {
```
gbalabasquer, line 95: Same here, add `emit Rely(msg.sender)` in the constructor.

src/dog.sol
```diff
@@ -82,6 +88,8 @@ contract Dog is LibNote {
       address clip,
       uint256 id
```
gbalabasquer, line 89: Wouldn't it be convenient to make the `id` indexed?

gbalabasquer: And what about logging the `tab`? Is it not useful?

gbalabasquer: Hmmm I'm actually seeing some collusions with the `Kick` event, which might be on purpose. And `tab` is in the `Kick` event so not actually necessary here.
I'm wondering if there was some analysis of which it is the best way to access all these values. In that way we could be sure we are adding the `indexed` keyword to the most correct values.

lucas-manuel: I added `indexed` to the `id` in `Bark`, we are copying the convention in cat for `Bite` with `Bark`:
```
event Bark(
      bytes32 indexed ilk,
      address indexed urn,
      uint256 ink,
      uint256 art,
      uint256 tab,
      address clip,
      uint256 indexed id
    );
```
Where we pass in `due` in for `tab`: `emit Bark(ilk, urn, dink, dart, due, milk.clip, id);` which I personally find kind of confusing so I'm open to changing it.

lucas-manuel: I remember we discussed that we should pass `due` instead of `tab` since `tab` is covered by `Kick`, but I think we should change it to `due` in the definition. (Or add `tab` as another param)

kmbarry1: Yeah I think calling it differently (`due` instead of `tab`) will help prevent confusion. There's already been some confusion with the present system (which logs a `tab` in both the Cat and Flipper, differing by the liquidation penalty).

src/clip.sol
kmbarry1: I don't feel incredibly strongly either, just thought it might be easier for human eyes when looking at Etherscan or something to lessen the chance of a mistake. I can live with it either way.

lucas-manuel:
Added `tab => due` change.

src/dog.sol
gbalabasquer: Looks good now.

gbalabasquer (approved):
LGTM.

lucas-manuel:
@gbalabasquer @kmbarry1 is this good to merge?

gbalabasquer:
I think is good to merge. There will be some minor changes in the logs when the other functions are finished but so far looks good.

PR #158, opened 2020-10-30
feat: Add Liquidation Price into `top` Calculation (SC-6715)

lucas-manuel:
Adding this PR to start a discussion about adding the liquidation price (`(art * rate / ink) * mat`) into the `top` calculation.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -66,8 +66,9 @@ contract Clipper is LibNote {
     AbacusLike public calc;  // Current price calculator
 
     uint256 public buf;   // Multiplicative factor to increase starting price  [ray]
-    uint256 public tail;  // Time elapsed before auction reset                 [seconds]
-    uint256 public cusp;  // Percentage drop before auction reset              [ray]
+    uint256 public tail;  // Required time elapsed before auction reset        [seconds]
+    uint256 public cusp;  // Required percentage drop before auction reset     [ray]
+    uint256 public lop;   // Percentage drop of previous top on auction reset  [ray]
 
     uint256   public kicks;   // Total auctions
     uint256[] public active;  // Array of active auction ids
@@ -144,8 +145,9 @@ contract Clipper is LibNote {
     // --- Administration ---
     function file(bytes32 what, uint256 data) external note auth {
         if      (what ==  "buf") buf  = data;
-        else if (what == "tail") tail = data; // Time elapsed    before auction reset
-        else if (what == "cusp") cusp = data; // Percentage drop before auction reset
+        else if (what == "tail") tail = data; // Required time elapsed    before auction reset
+        else if (what == "cusp") cusp = data; // Required percentage drop before auction reset
+        else if (what == "lop")  lop  = data; // Percentage drop of previous top on auction reset
         else revert("Clipper/file-unrecognized-param");
     }
     function file(bytes32 what, address data) external note auth {
@@ -227,7 +229,7 @@ contract Clipper is LibNote {
         (PipLike pip, ) = spot.ilks(ilk);
         (bytes32 val, bool has) = pip.peek();
         require(has, "Clipper/invalid-price");
-        sales[id].top = rmul(rdiv(mul(uint256(val), 10 ** 9), spot.par()), buf);
+        sales[id].top = max(rmul(sales[id].top, lop), rmul(rdiv(mul(uint256(val), BLN), spot.par()), buf));
 
         emit Redo(id, sales[id].top, sales[id].tab, sales[id].lot, sales[id].usr);
     }
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -117,8 +117,6 @@ contract Clipper is LibNote {
         address indexed usr
     );
 
-    event Debug(uint i, uint val);
-
     // --- Init ---
     constructor(address vat_, address spot_, address dog_, bytes32 ilk_) public {
         vat  = VatLike(vat_);
@@ -205,11 +203,6 @@ contract Clipper is LibNote {
         (PipLike pip, uint256 mat) = spot.ilks(ilk);
         (bytes32 val, bool has) = pip.peek();
         require(has, "Clipper/invalid-price");
-
-        emit Debug(1, lip);
-        emit Debug(2, rmul(lip, mat));
-        emit Debug(3, rdiv(mul(uint256(val), BLN), spot.par()));
-
         sales[id].top = rmul(max(rmul(lip, mat), rdiv(mul(uint256(val), BLN), spot.par())), buf);
 
         emit Kick(id, sales[id].top, tab, lot, usr);
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -101,7 +101,7 @@ contract Clipper is LibNote {
 
     event Take(
         uint256 indexed id, 
-        uint256 max,
+        uint256 cap,
         uint256 price,
         uint256 owe,
         uint256 tab,
@@ -235,7 +235,7 @@ contract Clipper is LibNote {
     // Buy amt of collateral from auction indexed by id
     function take(uint256 id,           // Auction id
                   uint256 amt,          // Upper limit on amount of collateral to buy  [wad]
-                  uint256 max,          // Maximum acceptable price (DAI / collateral) [ray]
+                  uint256 cap,          // Maximum acceptable price (DAI / collateral) [ray]
                   address who,          // Receiver of collateral, payer of DAI, and external call address
                   bytes calldata data   // Data to pass in external call; if length 0, no call is done
     ) external lock isStopped(2) {
@@ -250,7 +250,7 @@ contract Clipper is LibNote {
         require(sub(now, sale.tic) <= tail && rdiv(price, sale.top) >= cusp, "Clipper/needs-reset");
 
         // Ensure price is acceptable to buyer
-        require(max >= price, "Clipper/too-expensive");
+        require(cap >= price, "Clipper/too-expensive");
 
         // Purchase as much as possible, up to amt
         uint256 slice = min(sale.lot, amt);  // slice <= lot
@@ -297,7 +297,7 @@ contract Clipper is LibNote {
             sales[id].lot = sale.lot;
         }
 
-        emit Take(id, max, price, owe, sale.tab, sale.lot, sale.usr);
+        emit Take(id, cap, price, owe, sale.tab, sale.lot, sale.usr);
     }
 
     function _remove(uint256 id) internal {
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -117,6 +117,8 @@ contract Clipper is LibNote {
         address indexed usr
     );
 
+    event Debug(uint i, uint val);
+
     // --- Init ---
     constructor(address vat_, address spot_, address dog_, bytes32 ilk_) public {
         vat  = VatLike(vat_);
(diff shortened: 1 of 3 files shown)


src/clip.sol
```diff
@@ -196,10 +200,10 @@ contract Clipper is LibNote {
         // Could get this from rmul(Vat.ilks(ilk).spot, Spotter.mat()) instead,
         // but if mat has changed since the last poke, the resulting value will
         // be incorrect.
-        (PipLike pip, ) = spot.ilks(ilk);
+        (PipLike pip, uint256 mat) = spot.ilks(ilk);
```
kmbarry1, line 203: The "liquidation price" is just defined as `debt/collateral` at liquidation time, it shouldn't include `mat`.

kmbarry1: Although I can see the argument for correcting for it: it makes the price higher in the case of a phony oracle price; the tradeoff is that you risk setting the price too high in real declines. Deciding which is better (without being able to see the future) requires having beliefs about the likelihood and severity of various scenarios; this might be an excellent thing to have Gauntlet take a look at.

src/clip.sol
```diff
@@ -196,10 +200,10 @@ contract Clipper is LibNote {
...
         (bytes32 val, bool has) = pip.peek();
         require(has, "Clipper/invalid-price");
-        sales[id].top = rmul(rdiv(mul(uint256(val), BLN), spot.par()), buf);
+        sales[id].top = rmul(rdiv(max(rmul(lip, mat), mul(uint256(val), BLN)), spot.par()), buf);
```
kmbarry1, line 206: Btw only `mul(uint256(val), BLN)` should be divided by `par`, as `par` is a conversion factor from DAI to the references asset (i.e. `par` has units of `dollars / DAI`). `lip` is already `DAI / collateral` and `mat` is dimensionless.

lucas-manuel: I'll fix this 👍

src/clip.sol
lucas-manuel: My opinion is that it is better to have the higher of the two (`include mat`) because since this is just the initial price of the dutch auction, the price discovery can still occur later on after some decrease in the scenario of a sharp market decline. But yes that's a great point about Gauntet. I'm going to leave as is if no one has any major objections and we can revisit during their review. @kmbarry1 if you think its better that I take out `mat` before the Gauntlet review I can do that.

kmbarry1: Let's leave it in for now; at a minimum we should get an opinion from someone who works on risk. There's a call with Gauntlet tomorrow, I'll bring it up there.

lucas-manuel: Perfect 👍  sounds good.

src/dog.sol
```diff
@@ -176,6 +180,7 @@ contract Dog is LibNote {
             id = ClipperLike(milk.clip).kick({
                 tab: tab,
                 lot: dink,
+                lip: div(mul(art, rate), ink),
```
kmbarry1, line 183: I think in principle we could save some gas here by doing:
```
due / dink
```
`dink` is required to be non-zero above so we don't need the safe division, and `due / dink` should have approximately the same value as `art * rate / ink`. I don't see any additional precision concerns there under normal risk parameter conditions, given the `dust` checks.

kmbarry1:
It looks like there's sufficient coverage for both code paths in the tests, although I think it would be nice to add two new tests that explicitly document the intended behaviors--one for a steeply declined or zero OSM price that uses the liquidation price, and one for a case that should use the OSM price instead (e.g. an underwater Vault).

src/dog.sol
lucas-manuel: Added 👍

lucas-manuel:
Used `due / dink` for `lip` calc, added test for OSM liquidation price case.

src/dog.sol
```diff
@@ -105,6 +105,10 @@ contract Dog is LibNote {
     function mul(uint256 x, uint256 y) internal pure returns (uint256 z) {
         require(y == 0 || (z = x * y) / y == x);
     }
+    function div(uint256 x, uint256 y) internal pure returns (uint256 z) {
```
gbalabasquer, line 108: This can be removed now.
Btw why would we need this function? Dividing by zero already throws by default (at least in solidity, not 100% sure at EVM level).
I think older versions of solidity didn't catch it, but should have been solved long time ago.

src/clip.sol
```diff
@@ -162,6 +162,9 @@ contract Clipper is LibNote {
     function min(uint256 x, uint256 y) internal pure returns (uint256 z) {
         return x <= y ? x : y;
     }
+    function max(uint256 x, uint256 y) internal pure returns (uint256 z) {
```
gbalabasquer, line 165: This function gets shadowed (solidity warning) by another usage of the word `max` in `take` function. I would rename the argument of the function.

src/dog.sol
```diff
@@ -176,6 +180,7 @@ contract Dog is LibNote {
             id = ClipperLike(milk.clip).kick({
                 tab: tab,
                 lot: dink,
+                lip: due / dink,
```
gbalabasquer, line 183: I feel something is wrong with this math and the logic. Having a better collateralized CDP means a lower starting price. So let's say the feeds report a lower price than the real, a well collateralized CDP will be much more punished than one that's really underwater.
I feel the starting price should somehow be related to the inverse of the `due/dink`.

lucas-manuel:
Made requested changes by @gbalabasquer.

> I feel something is wrong with this math and the logic. Having a better collateralized CDP means a lower starting price. So let's say the feeds report a lower price than the real, a well collateralized CDP will be much more punished than one that's really underwater.
> I feel the starting price should somehow be related to the inverse of the `due/dink`.

This is an interesting point...

src/dog.sol
kmbarry1: There's no way for the auction contract to know, fundamentally, whether the feeds are wrong. The goal is to have some sort of fallback to a saner price if the feed price is super low. Even without this change, a very low oracle price punishes well-collateralized CDPs more (it's always worse to get liquidated if you have higher collateral/debt). Using liquidation price, corrected for `mat`, as a floor acts to make the worst-case less bad. Better to be liquidated at a low price than zero. I.e. this PR never makes the starting price any *lower* than it otherwise would, so it's still a strict improvement.

IMO it may be better to just invest in making the oracles more secure than trying to get extra clever here, although if you have any good ideas let's hear them :).

lucas-manuel: I agree with @kmbarry1 that in an oracle attack scenario this outcome is strictly better than not having `lip`. I'm okay with leaving as is.

gbalabasquer: Couldn't it happen that a very underwater vault won't be auctioned because the initial price is too high due this function?
I guess in that case it could be reset after the first cycle, which at least for now is not using this liquidation price.
Which btw, shouldn't we consider using it in the reset function somehow as well?

lucas-manuel: Yes so this is something I wanted to bring up. I wasn't sure if we wanted to use it in the reset function, because then the auction's `top`s will never decrease during reset which would be bad in a scenario of a steep market decline. However if there was an oracle attack and no one bid on a CDP that was really expensive for example (highly collateralized), then on reset that price would go to zero.

lucas-manuel: We could have another gov param that could be used to calculate the reset `top` using the previous `top`, decreasing it by a percentage.

gbalabasquer: Hmmm you mean highly underwater, right? As if the CDP is highly collateralized, the liquidation will start at a lower price.
But yes, to your point, not really sure what to do with `reset`, as the same logic than in the originally liquidation would apply here, however the full property of `reset` wouldn't work if we do this.
I think these are the reasons why I feel this functionality is a bit forced.

lucas-manuel: Yes sorry that's what I meant. And I do think that governance param solution could work for the reset, definitely open to other ideas though. I'll actually do a quick implementation of what I'm suggesting in the next commit.

kmbarry1: It might be interesting to do:
```
RESET_PRICE = max(buf*OSM_PRICE, PCT*top)
```
where `PCT` is the new gov parameter encoding the largest allowed (percentage-wise) decrease in `top`.

lucas-manuel: Thats actually what I just did! In the last commit. 

`sales[id].top = max(rmul(sales[id].top, lop), rmul(rdiv(mul(uint256(val), BLN), spot.par()), buf));`

I made `lop` directly multiply instead of doing `sub(RAY, lop)` (E.g. 25% decrease => `lop = ray(0.75 ether)`). I did this because if `lop` isn't set for whatever reason it will just be zero, so it won't be considered in the equation. This is IMO better that it working out to be `1` if its not set, which will just set it back to the old `top`.

kmbarry1: Yep, that's what I had in mind.

kmbarry1:
How do we feel here?

Pros:
 * get strictly higher prices if oracles are attacked and set too low
 * more predictable/controlled price decreases

Cons:
 * might set the starting price too high if a market decline has really occurred (and the error is worse the more undercollateralized a CDP is)
 * higher gas costs on every liquidation and auction reset
 * extra code complexity
 * another governance parameter
 * price might have even more trouble catching up in a rapidly falling market

Did I miss anything?

IMO it comes down to how likely you think an oracle attack is--we know it's a very severe outcome, but also unlikely, so is it worth the downsides above?

lucas-manuel:
> How do we feel here?
> 
> Pros:
> 
> * get strictly higher prices if oracles are attacked and set too low
> * more predictable/controlled price decreases
> 
> Cons:
> 
> * might set the starting price too high if a market decline has really occurred (and the error is worse the more undercollateralized a CDP is)
> * higher gas costs on every liquidation and auction reset
> * extra code complexity
> * another governance parameter
> * price might have even more trouble catching up in a rapidly falling market
> 
> Did I miss anything?
> 
> IMO it comes down to how likely you think an oracle attack is--we know it's a very severe outcome, but also unlikely, so is it worth the downsides above?

This might be over complicating things, but we could have an instant access module that could modify a boolean in the case of an oracle attack to use the liquidation price in that scenario:
```
sales[id].top = 
            oracleSafe ? 
            rmul(rdiv(mul(uint256(val), BLN), spot.par()), buf) : 
            rmul(max(rmul(lip, mat), rdiv(mul(uint256(val), BLN), spot.par())), buf);
```

However this still requires us to calculate and pass in `lip` on every `kick()`. IMO it might be worth that extra gas to have a safeguard against the system getting attacked, but I do understand that it's very unlikely. @gbalabasquer @kmbarry1 what do you guys think? If we think the cons outweigh the pros here I'm okay with closing this PR.

gbalabasquer:
> This might be over complicating things, but we could have an instant access module that could modify a boolean in the case of an oracle attack to use the liquidation price in that scenario:

Hmmm but how would that IAM would be executed and under which rules? I think at that point yes, we would be over complicating this too much.

> @gbalabasquer @kmbarry1 what do you guys think? If we think the cons outweigh the pros here I'm okay with closing this PR.

I would feel much better about it if we had a system that protects more to the cdps that are more collateralized than the exact opposite which it is doing now.
I know that a low initial price is better than 0, but I feel the cons list is too long for a functionality that IMO is a bit broken conceptually.
So in my personal opinion I'd vote for desisting on implementing it or would start thinking in a different angle to tackle it.

kmbarry1:
Definitely agree with Gonzalo that assuming an IAM on top is excessive, complexity-wise.

I'm a little more ambivalent on whether this is worth it. I guess it probably needs a more exhaustive analysis of how an oracle attack would be likely to play out (against one, several, or all collateral types) to figure out exactly how much help this would be. Which, given how close we are to freezing the code, means maybe we should put this aside for now and instead examine it as part of the audit process. If a clear case emerges for needing some mitigation, then we can revisit it.

Btw Gonzalo, I've tried to think about how to address your concern (the fact that it works worse for better-collateralized Vaults) and I think that could end up being very messy, without any trustworthy price oracle...

lucas-manuel:
@kmbarry1 so should we close this PR or leave it open you think (if we are going to reexamine during the audit process, which I agree is a good idea)?

kmbarry1:
> @kmbarry1 so should we close this PR or leave it open you think (if we are going to reexamine during the audit process, which I agree is a good idea)?

Hmmmm, maybe we'll close it for now so no one accidentally spends time reviewing when there are more important tasks? PRs can always be re-opened.

godsflaw:
closing for now

PR #161, opened 2020-11-03
feat: End Integration and `yank()` (SC-6352 + SC-6353)

lucas-manuel:


src/clip.sol
```diff
@@ -326,7 +326,10 @@ contract Clipper is LibNote {
     }
 
     // Cancel an auction during ES
-    function yank() external note auth {
-        // TODO
+    function yank(uint id) external auth {
```
kmbarry1, line 329: IMO we should go ahead and call `_remove(id)` here, since we might call this function as part of a spell after a circuit breaker trip, and otherwise we can't keep the list of active ids consistent.

src/end.sol
```diff
@@ -283,11 +312,29 @@ contract End is LibNote {
         tag[ilk] = wdiv(spot.par(), uint(pip.read()));
     }
 
+    function halt(bytes32 ilk, uint256 id) external note {
```
kmbarry1, line 315: Okay let the bikeshedding begin: I prefer `sqip`, or maybe `scip`, here myself.

src/end.sol
```diff
@@ -283,11 +312,29 @@ contract End is LibNote {
...
+        require(tag[ilk] != 0, "End/tag-ilk-not-defined");
+
+        (address _clip,,,) = dog.ilks(ilk);
+        ClipLike clip = ClipLike(_clip);
+        (, uint rate,,,) = vat.ilks(ilk);
+        (, uint256 tab, uint256 lot, address usr,,) = clip.sales(id);
+        
+        vat.suck(address(vow), address(vow),  tab);
+        vat.hope(address(clip)); // Should we have this?
```
kmbarry1, line 324: I don't see why this is needed. It's necessary in `skip` since there's a `move` call that transfer `dai` away from the caller, but `Clipper.yank` just `fluxes` collateral to the caller which doesn't require any special permission.

src/end.sol
lucas-manuel: I was thinking that, but my only issue is that since `sqip, scip and skip` are homonyms it might get confusing when we are discussing the functions verbally. If we want to maintain similarity with `skip` I'd personally prefer `swip`.

src/clip.sol
lucas-manuel: Fixed

src/end.sol
```diff
@@ -283,11 +312,28 @@ contract End is LibNote {
...
 
+    function halt(bytes32 ilk, uint256 id) external note {
+        require(tag[ilk] != 0, "End/tag-ilk-not-defined");
+
+        (address _clip,,,) = dog.ilks(ilk);
+        ClipLike clip = ClipLike(_clip);
+        (, uint rate,,,) = vat.ilks(ilk);
+        (, uint256 tab, uint256 lot, address usr,,) = clip.sales(id);
+        
+        vat.suck(address(vow), address(vow),  tab);
```
gbalabasquer, line 323: @kmbarry1 @lucas-manuel I know this same line is also in `skip` so there has to be some reason to be there but I can't get why (haven't done a deep analysis though).
Probably there is a an obvious reason I'm missing, but executing `vow.heal(tab)` after `skip` or `halt` reverts this action so no getting exactly what is the purpose. There is probably a scenario that I'm not considering where this is necessary.

src/end.sol
kmbarry1: Hmmmm okay fair, although I'd kind of think it would be either obvious from context or easy to clarify verbally.

In the homonym arena, I also quite like `skyp`. We could even pronounce it "skype" in verbal contexts to avoid ambiguity.

In the non-homonym arena, `snip` might be a little better mnemonically than `swip` (I certainly dislike `halt`, it gives the wrong impression IMO).

src/end.sol
kmbarry1: The `grab` call at the end is going to subtract `tab` from the `Vow's` `sin` balance. However, it's not guaranteed that the `Vow` will actually have that much `sin`, meaning you might get an underflow that reverts the transaction. This `suck` call ensures that the necessary `sin` is present in the `Vow's` balance, with equivalent `dai` so that the net effect on `Vat.dai(vow) - Vat.sin(vow)` is the same. And the reason you want to add to the system balance in the first place is to cancel out the `sin` that was assigned to the `Vow` when the position was liquidated. It's not an exact cancellation because it now includes the liquidation penalty; this is also by design, since the restored Vault also has the liquidation penalty added to its debt, meaning that extra collateral will be withheld by `skim` to cover that penalty, which now goes to DAI holders because it is in the surplus buffer.

gbalabasquer: Yes perfect, got it, that possible underflow is what I was missing. Thanks!

src/end.sol
lucas-manuel: I like `snip` actually, that's my fave so far. I'm also okay with `skyp`.

src/test/end.t.sol
```diff
@@ -483,6 +504,126 @@ contract EndTest is DSTest {
...
+        assertEq(vat.gem("gold", address(end)),         0); // From grab in end.snip() 
+        assertEq(vat.sin(address(vow)),       art1 * rate); // From grab in dog.bark()
+        assertEq(vat.vice(),                  art1 * rate); // From grab in dog.bark()
+        assertEq(vat.debt(),            tab + art1 * rate); // From frob and suck
+        assertEq(vat.dai(address(vow)),               tab); // From vat.suck()
+        assertEq(end.Art("gold") * rate,              tab); // Incrementing total Art in End
+
+        (uint ink3, uint art3) = vat.urns("gold", urn1);    // CDP after snip
+        assertEq(ink3, 10 ether);                           // All collateral returned to CDP
+        assertEq(art3 * rate, rad(16.5 ether));             // Tab amount of normalized debt transferred back into CDP
```
kmbarry1, line 565: I think I'd structure the assert as `assertEq(art3, tab / rate);`--you already assert on the value of tab, it's more readable, and remains correct even if `rate` doesn't evenly divide `tab`.

src/test/end.t.sol
```diff
@@ -483,6 +504,126 @@ contract EndTest is DSTest {
...
+        Usr ali = new Usr(vat, end);
+
+        // Make a CDP:
+        address urn1 = address(ali);
+        gold.gemA.join(urn1, 10 ether);
+        ali.frob("gold", urn1, urn1, urn1, 10 ether, 15 ether);
+        (uint ink1, uint art1) = vat.urns("gold", urn1); // CDP before liquidation
+        (, uint rate,,,) = vat.ilks("gold");
+        assertEq(vat.gem("gold", urn1), 0);
+        assertEq(rate, ray(1 ether));
```
kmbarry1, line 521: I get very nervous when we only test with `rate == RAY`...it makes it very easy for bugs stemming from an incorrect use of `rate` to slip by. Do you mind either changing this test to use a non-zero `rate`, or else adding another test that does so?

src/test/end.t.sol
```diff
@@ -483,6 +504,126 @@ contract EndTest is DSTest {
...
+        assertEq(vat.sin(address(vow)),       art1 * rate); // From grab in dog.bark()
+        assertEq(vat.vice(),                  art1 * rate); // From grab in dog.bark()
+        assertEq(vat.debt(),            tab + art1 * rate); // From frob and suck
+        assertEq(vat.dai(address(vow)),               tab); // From vat.suck()
+        assertEq(end.Art("gold") * rate,              tab); // Incrementing total Art in End
+
+        (uint ink3, uint art3) = vat.urns("gold", urn1);    // CDP after snip
+        assertEq(ink3, 10 ether);                           // All collateral returned to CDP
+        assertEq(art3 * rate, rad(16.5 ether));             // Tab amount of normalized debt transferred back into CDP
+
```
kmbarry1, line 566: IMO the rest of the test after this point really isn't necessary for verifying the `snip` functionality and can be omitted.

src/test/end.t.sol
```diff
@@ -511,22 +518,31 @@ contract EndTest is DSTest {
...
 
+        // To check this yourself, use the following rate calculation (example 8%):
+        //
+        // $ bc -l <<< 'scale=27; e( l(1.08)/(60 * 60 * 24 * 365) )'
+        uint256 EIGHT_PCT = 1000000002440418608258400030;
+        jug.file("gold", "duty", EIGHT_PCT);
+        hevm.warp(now + 10 days);
+        jug.drip("gold");
+        (, uint rate,,,) = vat.ilks("gold");
+
```
kmbarry1, line 529: Please use `vat.fold` directly to avoid depending on `Jug` unnecessarily.

src/test/end.t.sol
```diff
@@ -483,6 +504,70 @@ contract EndTest is DSTest {
...
+        assertEq(vat.gem("gold", address(gold.clip)), lot); // From grab in dog.bark()
+        assertEq(vat.sin(address(vow)),       art1 * rate); // From grab in dog.bark()
+        assertEq(vat.vice(),                  art1 * rate); // From grab in dog.bark()
+        assertEq(vat.debt(),                  art1 * rate); // From frob
+        assertEq(vat.dai(address(vow)),                 0); // vat.suck() hasn't been called 
+
+        end.snip("gold", id);
+
+        assertEq(dog.Dirt(),                            0); // From clip.yank()
```
kmbarry1, line 557: I'd add assertions on the Clipper state as well--`sales` entry cleared, and `id` removed from the list of active auctions. Otherwise, we have no test coverage for the state changes that `yank` is supposed to effect.

src/test/end.t.sol
```diff
@@ -483,6 +504,101 @@ contract EndTest is DSTest {
...
+        assertEq(vat.gem("gold", address(gold.clip)), lot1); // From grab in dog.bark()
+        assertEq(vat.sin(address(vow)),        art1 * rate); // From grab in dog.bark()
+        assertEq(vat.vice(),                   art1 * rate); // From grab in dog.bark()
+        assertEq(vat.debt(),                   art1 * rate); // From frob
+        assertEq(vat.dai(address(vow)),                  0); // vat.suck() hasn't been called 
+
+        end.snip("gold", id);
+
+        uint256 tab2;
+        uint256 lot2;
```
kmbarry1, line 572: Any reason not to move these inside the scoping block below?

src/test/end.t.sol
```diff
@@ -483,6 +504,101 @@ contract EndTest is DSTest {
...
+            (pos2, tab2, lot2, usr2, tic2, top2) = gold.clip.sales(id);  
+            assertEq(pos2,          0);
+            assertEq(tab2,          0); 
+            assertEq(lot2,          0);
+            assertEq(usr2,  address(0));
+            assertEq(uint256(tic2), 0);
+            assertEq(uint256(top2), 0);
+        }
+
+
```
kmbarry1, line 587:
```suggestion
```

src/test/end.t.sol
```diff
@@ -483,6 +504,101 @@ contract EndTest is DSTest {
...
+        {
+            address usr2;
+            uint96  tic2;
+            uint256 top2; 
+            uint256 pos2;
+            (pos2, tab2, lot2, usr2, tic2, top2) = gold.clip.sales(id);  
+            assertEq(pos2,          0);
+            assertEq(tab2,          0); 
+            assertEq(lot2,          0);
+            assertEq(usr2,  address(0));
```
kmbarry1, line 582:
```suggestion
            assertEq(usr2, address(0));
```

kmbarry1 (approved):
LGTM modulo some nits

I'd get approval from @gbalabasquer too before merging, as this is tricky & important logic.

src/test/end.t.sol
lucas-manuel: Good point I'll move them in, I copied this block from the first check where I needed `tab1` and `lot1` out of scope.

gbalabasquer (approved):
LGTM

PR #162, opened 2020-11-09
change price API to accept seconds elapsed since auction start

kmbarry1:
This should be a bit friendlier to integrators wanting to check what the price will be at some time in the future. It also makes more logical sense--the time elapsed since the start of the auction is all that matters, not the exact start time, so this is more explicit and thus clearer to readers of the code.

### DIFF
--- a/src/abaci.sol
+++ b/src/abaci.sol
@@ -15,14 +15,14 @@
 
 pragma solidity >=0.5.12;
 
-// interface Abacus {
-//     // 1st arg: initial price           [ray]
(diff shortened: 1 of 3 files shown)


lucas-manuel:
LGTM, should we implement the interface in this PR?

kmbarry1:
Depends on whether we're ready to switch the CI to use `0.6.7`.

lucas-manuel (approved):
Ok I'll approve and we can update it later.

PR #165, opened 2020-11-11
feat: Upgrade dog, clip to 0.6.7, Update CI (SC-7268)

lucas-manuel:
If anyone sees another opportunities to leverage 0.6.7 in these files (or the end) please let me know.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -55,13 +55,13 @@ contract Clipper {
     }
 
     // --- Data ---
-    bytes32 immutable public ilk;      // Collateral type of this Clipper
+    bytes32  immutable public ilk;   // Collateral type of this Clipper
+    VatLike  immutable public vat;   // Core CDP Engine
+    SpotLike immutable public spot;  // Collateral price module
 
-    address              public vow;   // Recipient of dai raised in auctions
-    VatLike    immutable public vat;   // Core CDP Engine
-    DogLike              public dog;   // Liquidation module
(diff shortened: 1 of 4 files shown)


src/clip.sol
```diff
@@ -55,13 +55,13 @@ contract Clipper {
...
-    address    public vow;   // Recipient of dai raised in auctions
-    VatLike    public vat;   // Core CDP Engine
-    DogLike    public dog;   // Liquidation module
-    SpotLike   public spot;  // Collateral price module
-    AbacusLike public calc;  // Current price calculator
+    address              public vow;   // Recipient of dai raised in auctions
+    VatLike    immutable public vat;   // Core CDP Engine
+    DogLike              public dog;   // Liquidation module
+    SpotLike   immutable public spot;  // Collateral price module
+    AbacusLike           public calc;  // Current price calculator
```
kmbarry1, line 64: Suggestion: separate `immutable` from mutable values for better organizatoin/clarity.

kmbarry1:
In abaci.sol, there's an interface that should be uncommented (with corresponding inheritance tags and `override` keywords that should also be uncommented). If you were planning to do that in another PR, that's fine too.

lucas-manuel:
@kmbarry1 I was thinking about adding that in another PR, but I think I'll just add it in this one.

kmbarry1 (approved):
LGTM

brianmcmichael (approved):

PR #167, opened 2020-11-16
refactor: Minor `dink` Gas Optimization

lucas-manuel:
Since `dart <= art`, we don't need to check `min` since `dink = ink` when `dart = art`.

Gas calculations:

```
360644 - full hole, old
360569 - full hole, new
    75 - gas savings
358067 - partial hole, old
357982 - partial hole, new
    85 - gas savings
```

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -177,7 +177,7 @@ contract Dog {
             require(dart == art || mul(art - dart, rate) >= dust, "Dog/leaves-dust");
         }
 
-        uint256 dink = min(ink, mul(ink, dart) / art);
+        uint256 dink = mul(ink, dart) / art;
 
         require(dink > 0, "Dog/null-auction");
         require(dart <= 2**255 && dink <= 2**255, "Dog/overflow");


kmbarry1 (approved):
LGTM

Nice catch!

gbalabasquer:
This thing comes from the old code: https://github.com/makerdao/dss/blob/b280b98e0c661d5d8effe2e9626e51978dab33ce/src/cat.sol#L128
where the `min` check was already there when in theory is not necessary.
When liq 1.2 came, we discussed that might have to do with some rounding issue possibility but we never prove it.
I'd suggest before removing this we make sure it never can happen that under any condition `dink` ends being higher than `ink`. cc @godsflaw

lucas-manuel:
> This thing comes from the old code:
> 
> https://github.com/makerdao/dss/blob/b280b98e0c661d5d8effe2e9626e51978dab33ce/src/cat.sol#L128
> 
> 
> where the `min` check was already there when in theory is not necessary.
> When liq 1.2 came, we discussed that might have to do with some rounding issue possibility but we never prove it.
> I'd suggest before removing this we make sure it never can happen that under any condition `dink` ends being higher than `ink`. cc @godsflaw

I remember this now. I'm trying to think of a scenario where `dink` could get greater than `ink`. The only situation where a rounding issue could come in is if `dart` and `art` are different, and the only time they differ is when `dart < art`, which would strictly translate to `dink < ink`. Since we are multiplying before dividing, it's not possible for there to be a rounding error when `dart = art`, so I think we are good here. `dink` should never be able to get above `ink`.

kmbarry1:
Lucas's assessment here is correct, the `min` is unnecessary since `dart <= art`.

gbalabasquer:
Yes I agree that should be fine, but the old code is the same thing, `lot <= ink` and it still did `art = min(art, mul(lot, art) / ink);`.
I'm just claiming I see the same case that we discussed before and we never reached a formal conclusion why was there.

kmbarry1:
> Yes I agree that should be fine, but the old code is the same thing, `lot <= ink` and it still did `art = min(art, mul(lot, art) / ink);`.
> I'm just claiming I see the same case that we discussed before and we never reached a formal conclusion why was there.

I personally don't recall ever discussing that. IMO it's just as obvious that the `min` is unnecessary there.

The only conspiracy theory I can come up with is maybe there was a concern that the division opcode in the EVM (or the `/` operator in Solidity) might start rounding differently at some point, which I don't think is terribly likely, given the amount of stuff that would break.

gbalabasquer:
Yeah, not sure if you were involved in that discussion.
I guess it was some un unnecessary operation that was wrongly introduced in this commit: https://github.com/makerdao/dss/commit/b5ad018ef0678c61f0de81dc82ea520b508f45a6#diff-b0cb825c61b97e7c1d56713799bc00f588ec6eee315da579658c64415dd5e41d
and then was never removed.

We could ask rain if there was something why it was done in this way, but as you say it was probably just wrongly added.

lucas-manuel:
> Yeah, not sure if you were involved in that discussion.
> I guess it was some un unnecessary operation that was wrongly introduced in this commit: [b5ad018#diff-b0cb825c61b97e7c1d56713799bc00f588ec6eee315da579658c64415dd5e41d](https://github.com/makerdao/dss/commit/b5ad018ef0678c61f0de81dc82ea520b508f45a6#diff-b0cb825c61b97e7c1d56713799bc00f588ec6eee315da579658c64415dd5e41d)
> and then was never removed.
> 
> We could ask rain if there was something why it was done in this way, but as you say it was probably just wrongly added.

I can't request @rainbreak's review, but @rainbreak do you see any issue with this PR?

lucas-manuel:
@kmbarry1 @gbalabasquer how do we feel about this PR?

kmbarry1:
My original LGTM still stands. Up to @gbalabasquer :)

gbalabasquer:
The change is the logic thing to do. I still have the doubt why originally was done in this way, but there is a high chance it was just a mistake.
So IMO it can be merged.

gbalabasquer:
Some finding using the symbolic testing from `dapptools`.

```
pragma solidity ^0.6.7;

import "ds-test/test.sol";

contract TestDsProveTest is DSTest {
    function setUp() public {

    }

    function prove_add(uint256 x, uint256 y) public {
        if (x + y < x) return; // no overflow
        assertTrue(x + y >= x);
    }

    function prove_mul(uint256 x, uint256 y) public {
        if (x * y < x) return; // no overflow
        assertTrue(x * y >= x);
    }

    function prove_mul2(uint256 x, uint256 y) public {
        if (x * y < x) return; // no overflow
        if (y == 0) return; // no 0 division
        assertTrue(x * y / y <= x);
    }
}
```

First two passed immediately, however third one has been running for an hour and still didn't get it to pass or fail.
I think that probably marks that the math result is not that simple as it seems.

And btw doing a slightly different testing:
```
function prove_mul2(uint256 x, uint256 y) public {
        if (x * y < x) return; // no overflow
        if (y == 0) return; // no 0 division
        assertTrue(x * y / y == x);
    }
```
I could get a counter example which provokes a lower value (not a problem for this discussed issue though).
These are the counter example values:
```
uint256 x = 62603589629148272772163168089776524384316105463266411221681349393467185877350;
uint256 y = 1941116169321835025403713069998527190438760697519441397794325847372366;
```

I'll update if it finds a couple of values for `x` and `y` that makes `x * y / y > x`.

If you want to play with it yourself, `hevm symbolic` can be used via `dapp test`. However it needs to be running the current `master` branch of `dapptools`, as it was merged after the latest release.

gbalabasquer:
Thinking it more carefully, finding a higher value would break the safe `mul` as it is describe as:
```
    }
```

That means, with a lower value, safe `mul` also breaks.
These specific 2 values should make the tx to revert:
```
uint256 x = 62603589629148272772163168089776524384316105463266411221681349393467185877350;
uint256 y = 1941116169321835025403713069998527190438760697519441397794325847372366;
```

gbalabasquer:
I've removed my previous comments, it was actually provoking an overflow with those values.

gbalabasquer (approved):

PR #169, opened 2020-11-18
feat: Add `needsRedo` Function (SC-7445)

lucas-manuel:


src/clip.sol
```diff
@@ -328,6 +328,17 @@ contract Clipper {
...
 
+    // Returns boolean for if an auction needs a redo (reset)
+    function needsRedo(uint256 id) external returns (bool) {
+        // Read auction data
+        Sale memory sale = sales[id];
+
+        // Compute current price [ray]
+        uint256 price = calc.price(sale.top, sub(now, sale.tic));
+
+        return (sale.tab > 0 && (sub(now, sale.tic) > tail || rdiv(price, sale.top) < cusp));
```
kmbarry1, line 339: WDYT about taking this line and turning it into an `internal` function which can be reused in all three places it occurs? This may slightly increase gas costs for `take` and `redo`, but since storage calls dominate gas costs anyway, IMO it's worth the readability/robustness.

lucas-manuel: Ill do a gas analysis on this today.

lucas-manuel:
@kmbarry1 just did a gas analysis of the latest commit:

```
test_auction_reset_cusp

  gas_needsRedo_current: 15975
  gas_redo_current: 36893
  gas_take_current: 143578
  
  gas_needsRedo_new: 16023
  gas_redo_new: 36981
  gas_take_new: 143660
  
  gas_needsRedo_diff: +48
  gas_redo_diff: -88
  gas_take_diff: -82
  
test_auction_reset_tail

  gas_needsRedo_current: 14950
  gas_redo_current: 35868
  gas_take_current: 143578

  gas_needsRedo_new: 14998
  gas_redo_new: 35956
  gas_take_new: 143660

  gas_needsRedo_diff: +48
  gas_redo_diff: -88
  gas_take_diff: -82
```

What are your thoughts?

kmbarry1:
My opinion: the cleaner, more obviously correct code is worth the very slight extra gas costs--if we end up compiling with optimizations on, this might go away anyway (and we can always change our minds about an internal implementation thing like this that doesn't affect the public API).

I might prefer `done` to `ripe` on the naming front.

kmbarry1:
Let's hear from @gbalabasquer as well.

gbalabasquer:
The internal `done` function looks good to me. However shouldn't `sale.tab > 0` be outside of this function? As it is actually necessary only for `needsRedo`.

lucas-manuel:
> The internal `done` function looks good to me. However shouldn't `sale.tab > 0` be outside of this function? As it is actually necessary only for `needsRedo`.

I added it because it was in the a separate require in both `redo` and `take`. On second thought though I will remove it because we should maintain that separate error message.

kmbarry1 (approved):
LGTM

src/clip.sol
```diff
@@ -328,6 +328,24 @@ contract Clipper {
         return active[idx];
     }
 
+    // Externally returns boolean for if an auction needs a redo
+    function needsRedo(uint256 id) external view returns (bool) {
+        // Read auction data
+        Sale memory sale = sales[id];
+
+        require(sale.tab > 0, "Clipper/not-running-auction");
```
gbalabasquer, line 336: I'm not sure this should `revert`.
I was imagining the return value as:
```
return sale.tab > 0 && done(sale, price);
```

lucas-manuel: Yes you're right I'll update this.

gbalabasquer (approved):
LGTM

kmbarry1 (approved):
LGTM

PR #170, opened 2020-11-30
fix: Input and state validation

kmbarry1:
Added input validation to `kick`; this is not strictly necessary, since the caller must be authorized, but is prudent for "future proofing" IMO and allows correctness of other logic in the contract to be guaranteed without depending on callers to be well-behaved.

I also regularized the check for whether or not an auction is running (`yank` was doing it differently and using a different mechanism)--I chose checking `sales[id].usr` to standardize on, but I could be talked into any equivalently good way of doing it.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -222,7 +222,7 @@ contract Clipper {
     function redo(uint256 id) external isStopped(2) {
         // Read auction data
         Sale memory sale = sales[id];
-        require(sale.tab > 0, "Clipper/not-running-auction");
+        require(sale.usr != address(0), "Clipper/not-running-auction");
 
         // Compute current price [ray]
         uint256 price = calc.price(sale.top, sub(now, sale.tic));
@@ -251,7 +251,7 @@ contract Clipper {
     ) external lock isStopped(2) {
         // Read auction data
         Sale memory sale = sales[id];
-        require(sale.tab > 0, "Clipper/not-running-auction");
+        require(sale.usr != address(0), "Clipper/not-running-auction");
 
         // Compute current price [ray]
         uint256 price = calc.price(sale.top, sub(now, sale.tic));
@@ -342,7 +342,7 @@ contract Clipper {
         // Compute current price [ray]
         uint256 price = calc.price(sale.top, sub(now, sale.tic));
 
-        return sale.tab > 0 && done(sale, price);
+        return sale.usr != address(0) && done(sale, price);
     }
 
     // Internally returns boolean for if an auction needs a redo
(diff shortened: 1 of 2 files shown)


kmbarry1:
Another thing I'm considering is adding a `lock` modifier to `yank`. Again it's not strictly necessary since the caller is trusted, but if for some weird reason a caller `yanked` from the external call in `take`, things could get...weird.

src/test/clip.t.sol
```diff
@@ -302,6 +302,27 @@ contract ClipperTest is DSTest {
...
+        assertTrue(!try_kick(0, 2 ether, address(1)));
+    }
+
+    function test_kick_zero_lot() public {
+        assertTrue(!try_kick(1 ether, 0, address(1)));
+    }
+
+    function test_kick_zero_usr() public {
+        assertTrue(!try_kick(1 ether, 2 ether, address(0)));
+    }
```
lucas-manuel, line 324: We should add this test for `take()` and `redo()` as well

lucas-manuel (changes requested):

lucas-manuel:
> Another thing I'm considering is adding a `lock` modifier to `yank`. Again it's not strictly necessary since the caller is trusted, but if for some weird reason a caller `yanked` from the external call in `take`, things could get...funky.

Tbh I don't see a scenario where this would be possible since the caller contract would have to be in `wards`.

kmbarry1:
> > Another thing I'm considering is adding a `lock` modifier to `yank`. Again it's not strictly necessary since the caller is trusted, but if for some weird reason a caller `yanked` from the external call in `take`, things could get...funky.
> 
> Tbh I don't see a scenario where this would be possible since the caller contract would have to be in `wards`.

Well, that's why I didn't add it...as the system stands now, it would be a bit redundant--the only caller is the `End` (and possibly a governance spell in extreme cases). But who knows what the future will bring.

src/test/clip.t.sol
kmbarry1: Done. Also converted a couple `testFail` tests for `redo` to be just `test` with different assertions (since that is less fragile).

lucas-manuel (approved):
LGTM.

brianmcmichael (approved):
Looks good

src/clip.sol
```diff
@@ -352,9 +358,10 @@ contract Clipper {
 
     // Cancel an auction during ES
     function yank(uint id) external auth {
-        require(sales[id].usr != address(0), "Clipper/usr-not-set");
-        dog.digs(ilk, sales[id].tab);
-        vat.flux(ilk, address(this), msg.sender, sales[id].lot);
+        Sale memory sale = sales[id];
```
gbalabasquer, line 361: I came a bit late with this. But it should be less gas efficient than before.
If I'm not wrong doing `Sale memory sale = sales[id];` will do 5 `SLOAD`s when before it was just doing 3.

kmbarry1: You are right, it is more gas efficient, I'll send another PR to fix it and add a test for `yank` that can be used to measure the gas in the future.

PR #172, opened 2020-12-04
Slither findings

kmbarry1:
Changes inspired by running [Slither](https://github.com/crytic/slither) against the code.


lucas-manuel (approved):
Wow I love Slither. LGTM.

gbalabasquer (approved):
It is a bit more difficult to reason about the `cut` value this way but saves gas. LGTM.

PR #173, opened 2020-12-09
Crytic config

kmbarry1:
Adds Crytic config; this is sufficient to get the tool working.

Only needs a single review IMO.


gbalabasquer (approved):
CI is failing but I think is actually doing what it should be doing if finds something. So LGTM.

kmbarry1:
Crytic is...no longer a thing, apparently. Closing.

PR #175, opened 2020-12-15
Clipper yank gas opt

kmbarry1:
A more gas-efficient implementation of `yank` (after I made it worse in a previous PR :P).

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -356,12 +356,11 @@ contract Clipper {
         emit SetBreaker(level);
     }
 
-    // Cancel an auction during ES
+    // Cancel an auction during ES or via governance action.
     function yank(uint id) external auth {
-        Sale memory sale = sales[id];
-        require(sale.usr != address(0), "Clipper/not-running-auction");
-        dog.digs(ilk, sale.tab);
-        vat.flux(ilk, address(this), msg.sender, sale.lot);
+        require(sales[id].usr != address(0), "Clipper/not-running-auction");
+        dog.digs(ilk, sales[id].tab);
+        vat.flux(ilk, address(this), msg.sender, sales[id].lot);
         _remove(id);
         emit Yank();
     }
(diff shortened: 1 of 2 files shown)


gbalabasquer (approved):
LGTM

src/clip.sol
```diff
@@ -356,12 +356,11 @@ contract Clipper {
...
-    // Cancel an auction during ES
+    // Cancel an auction during ES or via governance action.
     function yank(uint id) external auth {
-        Sale memory sale = sales[id];
-        require(sale.usr != address(0), "Clipper/not-running-auction");
-        dog.digs(ilk, sale.tab);
-        vat.flux(ilk, address(this), msg.sender, sale.lot);
+        require(sales[id].usr != address(0), "Clipper/not-running-auction");
+        dog.digs(ilk, sales[id].tab);
+        vat.flux(ilk, address(this), msg.sender, sales[id].lot);
```
rainbreak, line 363: curious if you tested the gas difference using the optimizer with a lot of runs?

kmbarry1: Well, I hadn't, but since you ask (using `solc` `0.6.7`, with the `note` modifier bodies commented out since you can't compile with optimizations on if your code contains explicit `mstore` calls):

10^6 optimizer runs:
...whole struct read: 36,064
...per-field reads: 34,210

No optimization:
...whole struct read: 41,005
...per-fields reads: 38,296

So the optimizer does succeed in narrowing the gap, and (optimizations+whole struct read) is better than (no optimizations+per-field reads) but per-field reads is still better than a whole-struct read when optimizations are on. Btw bumping the runs up to 10^9 doesn't change the numbers, so likely the numbers above are already the limit of the current optimizer's abilities.

kmbarry1:
Heads up: I'm going to merge this later today if there are no objections.

PR #179, opened 2021-01-13
Add remaining art to dart to prevent unliquidatable vaults

brianmcmichael:
Current situation allows for a large vault to become unliquidatable if it leaves a small dusty amount. 

This adds the dusty amount to the dart to liquidate the entire amount.

This comes at the cost of exceeding the hole, but it prevents leaving a dusty amount and prevents locking the collateral.

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -173,15 +173,10 @@ contract Dog {
 
             dart = min(art, mul(room, WAD) / rate / milk.chop);
 
-            uint256 mart = mul(art - dart, rate);
-
-            if (mart < dust) {
+            if (mul(art - dart, rate) < dust) {
                 // avoid leaving a dusty vault to prevent unliquidatable vaults
-                dart = add(dart, mart);
+                dart = art;
             }
-
-            // Verify that CDP is not left in a dusty state
-            require(dart == art || mart >= dust, "Dog/leaves-dust");
         }
 
         uint256 dink = mul(ink, dart) / art;
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -181,7 +181,7 @@ contract Dog {
             }
 
             // Verify that CDP is not left in a dusty state
-            require(dart == art || mul(art - dart, rate) >= dust, "Dog/leaves-dust");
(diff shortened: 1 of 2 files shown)


src/dog.sol
```diff
@@ -173,8 +173,15 @@ contract Dog {
 
             dart = min(art, mul(room, WAD) / rate / milk.chop);
 
+            uint256 mart = mul(art - dart, rate);
+
+            if (mart < dust) {
+                // avoid leaving a dusty vault to prevent unliquidatable vaults
+                dart = add(dart, mart);
```
gbalabasquer, line 180: I guess it is easier to say `dart = art;`

src/dog.sol
```diff
@@ -173,8 +173,15 @@ contract Dog {
 
             dart = min(art, mul(room, WAD) / rate / milk.chop);
 
+            uint256 mart = mul(art - dart, rate);
```
gbalabasquer, line 176: This actually should be renamed by something else. `art` is used for normalized debt, in this case as we are multiplying by `rate` it is not actually an `art`.

src/dog.sol
```diff
@@ -173,8 +173,15 @@ contract Dog {
...
+
+            if (mart < dust) {
+                // avoid leaving a dusty vault to prevent unliquidatable vaults
+                dart = add(dart, mart);
+            }
+
             // Verify that CDP is not left in a dusty state
-            require(dart == art || mul(art - dart, rate) >= dust, "Dog/leaves-dust");
+            require(dart == art || mart >= dust, "Dog/leaves-dust");
```
gbalabasquer, line 184: Actually this `require` is not needed anymore, as if `mart` is lower than `dust` will be taken care in the prev step. So I think the whole thing can be refactor to this:

gbalabasquer:
```
if (mul(art - dart, rate) < dust) {
    // avoid leaving a dusty vault to prevent unliquidatable vaults
    dart = art;
}
```
(removing the `require` and `uint256 mart = mul(art - dart, rate);`)

godsflaw (approved):
LGTM, deal with TODO after

PR #180, opened 2021-01-13
Add extra function to pay rewards to external address

brianmcmichael:
Overloads the bark function so that a keeper can send the rewards to a different address.

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -150,7 +150,7 @@ contract Dog {
     function chop(bytes32 ilk) external view returns (uint256) { return ilks[ilk].chop; }
 
     // --- CDP Liquidation: all bark and no bite ---
-    function bark(bytes32 ilk, address urn, address guy) public returns (uint256 id) {
+    function bark(bytes32 ilk, address urn, address guy) external returns (uint256 id) {
         require(live == 1, "Dog/not-live");
 
         (uint256 ink, uint256 art) = vat.urns(ilk, urn);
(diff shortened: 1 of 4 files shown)


src/dog.sol
```diff
@@ -202,11 +202,15 @@ contract Dog {
             });
         }
 
-        vat.suck(address(vow), msg.sender, add(milk.tip, wmul(due, milk.chip)));
+        vat.suck(address(vow), guy, add(milk.tip, wmul(due, milk.chip)));
 
         emit Bark(ilk, urn, dink, dart, due, milk.clip, id);
     }
 
+    function bark(bytes32 ilk, address urn) external returns (uint256 id) {
```
gbalabasquer, line 210: I'd tend to prefer to not have this function directly if we want to have a different address for the `tip`.
So only keep `bark(bytes32 ilk, address urn, address guy)` and make it `external`.

src/dog.sol
```diff
@@ -150,7 +150,7 @@ contract Dog {
     function chop(bytes32 ilk) external view returns (uint256) { return ilks[ilk].chop; }
 
     // --- CDP Liquidation: all bark and no bite ---
-    function bark(bytes32 ilk, address urn) external returns (uint256 id) {
+    function bark(bytes32 ilk, address urn, address guy) public returns (uint256 id) {
```
gbalabasquer, line 153: I think we should use `usr` which is the standard now.

brianmcmichael: fixed

src/dog.sol
brianmcmichael: removed the overload

src/dog.sol
gbalabasquer: I think you missed to push the change or something is not getting updated.

gbalabasquer (approved):
LGTM

godsflaw (approved):
LGTM, remember to add this to the redo() when we add incentives there too.

PR #181, opened 2021-01-14
LIQ-2.0 20210114 updates

godsflaw:
- updates some comments with
- fix Makefile
- adds missing events
- `now` to `block.timestamp`
- TODO
- [x] add yank id
- [ ] add incentive like `bark()` in `redo()`
- [ ] check the possibility of 1 wei rounding error
- [x] consider circuit breaking flash loans

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,3 @@
 build  :; dapp --use solc:0.6.11 build
 clean  :; dapp clean
-test   :; dapp --use solc:0.6.11 test
+test   :; dapp --use solc:0.6.11 test -v
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,3 @@
-remappings=ds-test/=lib/ds-test/src/ ds-token/=lib/ds-token/src/ ds-note/=lib/ds-token/lib/ds-stop/lib/ds-note/src/ ds-value/=lib/ds-value/src/ erc20/=lib/ds-token/lib/erc20/src/ ds-math/=lib/ds-token/lib/ds-math/src/ ds-stop/=lib/ds-token/lib/ds-stop/src/ ds-thing/=lib/ds-value/lib/ds-thing/src/ ds-auth/=lib/ds-token/lib/ds-stop/lib/ds-auth/src/
-opts=--combined-json=abi,bin,bin-runtime,srcmap,srcmap-runtime,ast
-files=$$(find src -type f -name '*.sol')
-
-ls:
-	echo ${files}
-
-clean:
-	rm -rf out/
-
-build: clean
-	mkdir -p out/
-	solc --overwrite ${remappings} ${opts} /=/ ${files} > out/dss.json
-
-test: build
-	hevm dapp-test --json-file out/dss.json
+build  :; dapp --use solc:0.6.11 build
+clean  :; dapp clean
+test   :; dapp --use solc:0.6.11 test
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -120,7 +120,7 @@ contract Clipper {
     );
 
     event SetBreaker(uint256 level);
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -1,16 +1,3 @@
...
-
-build: clean
-	mkdir -p out/
-	solc --overwrite ${remappings} ${opts} /=/ ${files} > out/dss.json
-
-test: build
-	hevm dapp-test --json-file out/dss.json
+build  :; dapp --use solc:0.6.11 build
+clean  :; dapp clean
+test   :; dapp --use solc:0.6.11 test
```
brianmcmichael, line 3: Add -v here.

src/clip.sol
```diff
@@ -362,6 +367,7 @@ contract Clipper {
         dog.digs(ilk, sales[id].tab);
         vat.flux(ilk, address(this), msg.sender, sales[id].lot);
         _remove(id);
+        // TODO: add id?
```
brianmcmichael, line 370: I think this is a yes. Event without the parameter doesn't seem useful.

src/clip.sol
```diff
@@ -272,6 +274,7 @@ contract Clipper {
         if (owe >= sale.tab) {
             owe = sale.tab;       // owe' <= owe
             slice = owe / price;  // Adjust slice; slice' = owe' / price <= owe / price == slice <= lot
+            // TODO check that this doesn't violate max from user
```
e18r, line 277: It doesn't since `owe'` is now lower or equal than its original value:

    slice' = owe' / price // from line 275 we know that owe' = sale.tab
    slice' = sale.tab / price // from line 274 we know that owe >= sale.tab
    slice' <= owe / price // from line 269 we know that owe = slice * price
    slice' <= (slice * price) / price
    slice' <= slice

src/clip.sol
gbalabasquer: Yeah, it should have the `id`.

src/clip.sol
gbalabasquer: Yeah it shouldn't violate the max price as `slice` is actually defined as `owe / price` as Emilio said.
Btw this is inside the function we might restructure.

src/clip.sol
```diff
@@ -281,12 +284,14 @@ contract Clipper {
...
 
         // Calculate remaining lot after operation
+        // TODO: could we end up with 1 wei rounding error?
         sale.lot = sale.lot - slice;  // safe because: slice <= lot
 
         // Send collateral to who
         vat.flux(ilk, address(this), who, slice);
 
         // Do external call (if defined)
+        // TODO: do we want to circuit break flash loans?
```
gbalabasquer, line 294: I feel that we shouldn't do it, but not really opposed to have it if the rest prefers so.

e18r: I'm working on it already: https://github.com/makerdao/dss/pull/182

Why not? I think it is by far the riskiest part of the whole code. Also it has zero unit tests rn.

I would even go as far as to suggest to launch liquidations 2.0 with external calls disabled, and only enable them when we make sure everything else works correctly.

gbalabasquer: It's one of the most interesting functionalities of auctions 2.0, so I wouldn't leave it out for the launch at all.
Technically any part of the code might have a bug and we are not adding functionality to circuit break them.
I agree that external calls are riskier but I'd say we better put the effort to make sure it is working fine and malicious calls attempts can't do anything to it.

e18r: @gbalabasquer OK. Instead I added tests for external calls: https://github.com/makerdao/dss/pull/185

brianmcmichael (approved):

gbalabasquer (approved):

PR #183, opened 2021-01-18
Liq 2.0 take impersonation bug

e18r:
Anybody can call `take` on behalf of anybody else. This is a bug.

As discussed with @gbalabasquer, there are at least two possible solutions:

1. Take the DAI from `msg.sender` instead of `who` by replacing it in `vat.move`. This is the simplest solution, but it puts an additional burden in the `who.clipperCall` function created by the caller, since it has to transfer the owed DAI to `msg.sender`.
2. Implement a `hope/nope/wish/can` permission delegation system in the Clipper similar to the one used by the Vat. This way, the caller would need to `clip.hope` their `who` contract before calling `take`. This solution, however, puts an additional burden in the Clipper since it now has to store an additional mapping.

Note: in this PR I added a test that checks this behavior. Since it is unexpected, the test fails, causing the CI check to fail.

### DIFF
--- a/src/test/clip.t.sol
+++ b/src/test/clip.t.sol
@@ -985,4 +985,15 @@ contract ClipperTest is DSTest {
         // Assert transfer of gem.
         assertEq(vat.gem(ilk, address(this)), preGemBalance + origLot);
     }
+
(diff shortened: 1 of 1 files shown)


e18r:
Closing in favor of #187

PR #185, opened 2021-01-18
Liq 2.0: add tests for external calls

e18r:
Add tests for `clip.take` using external calls.

* `testFail_not_enough_dai` tests that a user without enough DAI can't complete the call.
* `test_flashsale` tests that a user without enough DAI can complete the call if they use an external call to get DAI.
* `testFail_reentrancy` tests that a user cannot call `clip.take` again from within the `who.clipperCall` function.
* `testFail_redo` tests that a use cannot call `clip.redo` from within the `who.clipperCall` function.

### DIFF
--- a/src/test/clip.t.sol
+++ b/src/test/clip.t.sol
@@ -73,6 +73,16 @@ contract BadGuy is Guy {
     }
 }
 
+contract RedoGuy is Guy {
+
+    constructor(Clipper clip_) Guy(clip_) public {}
+
+    function clipperCall(uint256 owe, uint256 slice, bytes calldata data)
+        external override {
+        clip.redo(1);
+    }
+}
+
 contract ClipperTest is DSTest {
     Hevm hevm;
 
@@ -90,6 +100,7 @@ contract ClipperTest is DSTest {
     address bob;
     address che;
     address dan;
+    address emi;
 
     uint256 WAD = 10 ** 18;
     uint256 RAY = 10 ** 27;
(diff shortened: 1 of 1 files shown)


gbalabasquer:
I think the spacing was already solved in another PR, so this will be probably bringing a lot of conflicts. Can you revert that and leave this pr just changing the main functionality?
If the spacing is not solved 100% with the other pr, we can take care of it later.

src/test/clip.t.sol
```diff
@@ -49,6 +49,28 @@ contract Guy {
     function bark(Dog dog, bytes32 ilk, address urn) external {
         dog.bark(ilk, urn);
     }
+
+    function clipperCall(uint256 owe, uint256 slice, bytes calldata data)
+        external virtual {
+        Vat(address(clip.vat())).suck(address(0), address(this), owe);
```
gbalabasquer, line 55: The test works, but maybe as an actual example I'd go for something closer to the reality where you can actually test that you are receiving the collateral up front.
So we'd need a contract that exchanges your collateral for DAI and you can leave a small profit to show the case of why someone go run it. It'd be a plain arbitrage operation.

src/test/clip.t.sol
```diff
@@ -903,4 +933,36 @@ contract ClipperTest is DSTest {
...
+            id:  1,
+            amt: 25 ether,
+            max: ray(5 ether),
+            who: address(che),
+            data: "hey"
+        });
+        assertEq(vat.dai(che), 0);
+    }
+
+    function testFail_reentrancy() public takeSetup {
```
gbalabasquer, line 959: You could test a second one trying to enter to `redo`.

src/test/clip.t.sol
e18r: Updated with a more realistic example

src/test/clip.t.sol
e18r: Added a `redo` test.

src/test/clip.t.sol
```diff
@@ -19,6 +20,82 @@ interface Hevm {
...
+    Exchange uniswap;
+
+    constructor(
+        Clipper clip_,
+        Vat vat_,
+        DSToken gold_,
+        GemJoin goldJoin_,
+        DSToken dai_,
+        DaiJoin daiJoin_,
+        Exchange uniswap_
```
gbalabasquer, line 59: Let's avoid third party names.

e18r: done.

gbalabasquer (approved):
LGTM

godsflaw (approved):

PR #187, opened 2021-01-20
Block take impersonation and implement approval

brianmcmichael:
Add a simple approval system for takes. Assumes `take()` passes itself as a `who`, otherwise require `who` is an approved receiver. Adds test of approval.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -70,6 +70,13 @@ contract Clipper {
     uint256   public kicks;   // Total auctions
     uint256[] public active;  // Array of active auction ids
 
+    // Manage take approvals
+    mapping(
+        address => mapping(
+            address => bool
+        )
+    ) public can;           // Owner => Allowed Owner => True/False
+
     struct Sale {
         uint256 pos;  // Index in active array
         uint256 tab;  // Dai to raise       [rad]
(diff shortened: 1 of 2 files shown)


brianmcmichael:
This pr cherry-picked b0a3221f9561664d31554b79eede07dad972f72e from #183

e18r:
This is a good approach.

I think we should follow the standard of `hope/nope/wish/can`. Having multiple standards for the same thing can become very confusing to our users. For instance, before calling `take`, our advanced users would have to call

    vat.hope(clip)
    clip.approve(who)

I think it's much more intuitive to call

    vat.hope(clip)
    clip.hope(who)

gbalabasquer:
Don't you guys think it would be much easier just doing the following changes:
```
vat.move(msg.sender, vow, owe);
```
and
```
ClipperCallee(who).clipperCall(owe, slice, msg.sender, data);
````

With just that we avoid this problem at all.
The only thing is `who.clipperCall` will need to do:
```
daiJoin.exit(sender, amount)
```

brianmcmichael:
@e18r this is a similar pattern to what we're using in https://github.com/makerdao/dss-cdp-manager

@gbalabasquer I'll try it.

brianmcmichael:
So we may want to consider moving toward [ERC-3156](https://eips.ethereum.org/EIPS/eip-3156) conformity here to solve this problem.

brianmcmichael:
Alternate solution here https://github.com/makerdao/dss/pull/189

gbalabasquer:
Closing this PR in favor of #189

PR #188, opened 2021-01-20
fix: move incentive to kick() in redo() (SC-8390)

godsflaw:
This fixes a bug where we missed incentives in `redo()` by moving incentives into the `clip.kick()` and `clip.redo()`.  The variables `chip` and `tip` were also moved to each ilk's respective `Clipper`.

The biggest methodology change is that rather than calculating the `tip` and `chip` on the normalized debt owed, we are calculating it on the entire `tab`.  Calculating these values on the normalized debt in `kick()` would require passing the normalized debt to `kick()` and then storing it.  We would then need to store the original `tab` in order to calculate the ratio of initial tab to partial tab later in the `redo()` so that we could only incentivize the `redo()` on the leftover partial liquidation.

Rather than doing all of that, I'm suggesting we simply use the `tab` to calculate incentives.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -60,6 +60,7 @@ contract Clipper {
     VatLike  immutable public vat;   // Core CDP Engine
     SpotLike immutable public spot;  // Collateral price module
 
+    // TODO: do we want to make vow and dog immutable?
     address    public vow;   // Recipient of dai raised in auctions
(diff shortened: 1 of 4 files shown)


gbalabasquer:
I agree this is the only simple way to do it, so IMO you can move on.

gbalabasquer:
Merging, it's pending some more tests.

PR #189, opened 2021-01-20
Prevent take impersonation.

brianmcmichael:
Alternative solution to preventing take impersonation.

I just realized that there are no tests of a `clipperCall()` implementation, so we will need to add one depending on the implementation we go with.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -70,6 +70,13 @@ contract Clipper {
     uint256   public kicks;   // Total auctions
     uint256[] public active;  // Array of active auction ids
 
+    // Manage take approvals
+    mapping(
+        address => mapping(
+            address => bool
+        )
+    ) public can;           // Owner => Allowed Owner => True/False
+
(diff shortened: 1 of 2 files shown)


gbalabasquer:
@brianmcmichael Emilio worked in some test on https://github.com/makerdao/dss/pull/185

gbalabasquer:
I think we should go with this solution. Much simpler than having to add an allowance mapping and its functions to the `Clipper`.
Should we merge the tests of #185 in this PR, adapt them and review the whole thing?

src/clip.sol
```diff
@@ -253,6 +254,7 @@ contract Clipper {
                   address who,          // Receiver of collateral, payer of DAI, and external call address
```
e18r, line 254: ~payer of DAI~

brianmcmichael: fixed

godsflaw (approved):

PR #190, opened 2021-01-20
Instead of ending auction if tab < dust, recalculate amounts

gbalabasquer:


gbalabasquer:
Still thinking if everything is correct and maybe I'm adding some extra test/s. But consider is good enough to start receiving reviews.

e18r (approved):
This is brilliant. You effectively managed to remove the dust limitation, which means it won't be necessary to refactor the whole function in order to make this dust calculation easier for the user, as we were considering.

After exhaustively reviewing every line, I think your code is flawless. However, to me at least it was hard to follow. That's why I decided to write [my own version](https://github.com/makerdao/dss/pull/191). It's exactly the same code, except it takes less lines and (to me at least) is easier to understand. Feel free to switch to that branch if you like it.

src/clip.sol
```diff
@@ -284,17 +284,23 @@ contract Clipper {
...
+                // Adjust slice
+                slice = owe / price;        // slice' = owe' / price <= owe / price == slice <= lot
+            } else if (slice < lot) {
+                // if slice == lot => auction completed => dust doesn't matter
                 (,,,, uint256 dust) = vat.ilks(ilk);
-                require(tab > dust, "Clipper/no-partial-purchase");
-                if (owe > tab - dust) { // owe would leave a dusty amount
-                    owe = tab - dust; // owe' <= owe
+                if (tab - owe < dust) {     // safe as owe < tab
+                    require(tab > dust, "Clipper/no-partial-purchase");
```
e18r, line 294: This `require` can safely go outside of its surrounding `if`. Not only would this save a bit of gas in some failure cases, but also I think it makes the code more understandable since what we're testing here has nothing to do with `tab - owe` being less than `dust`. We already know that `owe < tab` (aka partial purchase) so we want to make sure we're not in that special case where `tab == dust`. For the sake of future maintainers, I would also add a comment:

    // if tab == dust, buyers have to buy the whole thing

gbalabasquer: I have some doubts about this check, after pushing I had pending to review it again.
Let's put an example. `tab = 10000`, `owe=7000`, `dust=2000`, `slice < lot`. Then enters to the `else if` condition.
Then the condition `10000 - 7000 < 2000 == false`, so you do not have to do any operation, why do you want to execute the `require` anyway?

gbalabasquer: The `tab > dust` was introduced as a better way instead of doing `sub(tab, dust)`. First because if someone gets a revert from it will get a specific message and second because it excludes the case where `tab == dust`, meaning that `owe` will be 0 and the whole operation is pointless.

gbalabasquer: So basically I considered the check should be inside the `if` because is when it actually matters, if you do not enter the `if` don't see the need to actually check it.

gbalabasquer: I'm still reviewing if the `require` is actually necessary at all.

PR #191, opened 2021-01-21
Liq 2.0 recalculate dust

e18r:
This is a simplification of @gbalabasquer's pull request #190.


gbalabasquer:
This simplification looks very good, I'm taking it to my PR, but there are just 2 minor things I may modify:
- There is one case that `slice` doesn't change, so basically is doing one extra operation in a case that is not necessary (waste of gas and maybe even having a rounding error when is not needed). So I say that one should be duplicated in the two parts of the code where it corresponds.
- With the calculation of the remaining `tab`, same thing but without the rounding problem. As the operation is not gas expensive probably the best to leave as you did it.

e18r:
Closing as @gbalabasquer incorporated these changes into his PR.

PR #193, opened 2021-01-22
Change needsRedo function to getStatus which also includes the price

gbalabasquer:


src/clip.sol
```diff
@@ -390,15 +390,15 @@ contract Clipper {
     }
 
     // Externally returns boolean for if an auction needs a redo
```
e18r, line 392: update comment

e18r (approved):

src/clip.sol
gbalabasquer: Fixed!

godsflaw (approved):

PR #194, opened 2021-01-22
Make vow immutable in the dog and vow and dog in clipper + check who …

gbalabasquer:
…is not dog in take external call


gbalabasquer:
Making `vow` and `dog` immutable makes that the whole thing needs to be redeployed if some of these contracts change. But consider that for example the `spotter` is already immutable and that one will probably have to be upgraded if for example we need to restructure oracles.

brianmcmichael (approved):

gbalabasquer:
We are closing this PR as we prefer to not have immutable variables for `vow` and `spot`

PR #196, opened 2021-01-25
Make chip and tip public + more tests

gbalabasquer:


brianmcmichael (approved):
++ these definitely need to be public.

Tests look good, too.

src/test/clip.t.sol
```diff
@@ -1138,6 +1173,65 @@ contract ClipperTest is DSTest {
...
+        clip.redo(1, address(345));
+        assertEq(vat.dai(address(345)), clip.chip() * tab / WAD);
+
+        vat.file(ilk, "dust", rad(110 ether) + 1); // 1 wei > than $110 (tab) as dust
+
+        hevm.warp(now + 300);
+        clip.redo(1, address(456));
+        assertEq(vat.dai(address(456)), 0);
+
+        vat.file(ilk, "dust", rad(20 ether)); // $20 dust
```
e18r, line 1208: add comment here explaining that you are testing the dusty lot case

e18r (approved):
LGTM

src/test/clip.t.sol
gbalabasquer: done!

src/test/clip.t.sol
```diff
@@ -1138,6 +1173,65 @@ contract ClipperTest is DSTest {
         assertTrue(try_redo(1, address(this)));
     }
 
+    function test_redo_incentive() public takeSetup {
+        // (, uint256 price) = clip.getStatus(1);
+        // assertEq(price, ray(5 ether));
```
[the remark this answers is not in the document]
gbalabasquer: Done!

PR #197, opened 2021-01-25
dog.bark: simplify calculation of dart

e18r:
Since dust is now a factor in the calculation of `dart`, this PR makes this process more straightforward by merging two calculations into one. Basically, in order to liquidate a whole vault, it must be true that

    room / chop > tab - dust

That is, the chop-adjusted room has to be bigger than the vault, excluding its dust value.

While maintaining similar gas expenditures, this PR not only makes the code more maintainable, but also fails to create a liquidation if the vault is unliquidatable.

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -120,6 +120,9 @@ contract Dog {
     function mul(uint256 x, uint256 y) internal pure returns (uint256 z) {
         require(y == 0 || (z = x * y) / y == x);
     }
+    function wdiv(uint x, uint y) internal pure returns (uint z) {
+        z = mul(x, WAD) / y;
+    }
 
     // --- Administration ---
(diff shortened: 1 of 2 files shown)


gbalabasquer:
We want to liquidate all vaults that fall under the liqudation price, even if the `dust` limit has increased over some of them.
So the conceptual error is thinking on "unliquidatable vaults" as there shouldn't be any.

gbalabasquer:
I mean with `unliquidatable vaults` the comment tries to say that because of the `hole` we do not want to leave vaults that nobody wants to liquidate. However the system shouldn't block to liquidate any vault.
What maybe we should consider is adding the dusty check in order to not pay the incentive if they are dusty values (just we did in `redo`). An altruistic bot could take them.

PR #198, opened 2021-02-01
Liq 2.0 formulas

e18r:


src/clip.sol
```diff
@@ -275,7 +275,77 @@ contract Clipper {
...
     }
 
-    // Buy amt of collateral from auction indexed by id
+    // Buy `amt` of collateral from auction indexed by `id`
+    //
+    // This function computes `owe`, i.e. the amount of DAI needed to purchase
+    // `amt` of collateral at the current price, as determined by the auction's
+    // abacus:
+    //
+    //     owe = amt * abacus.price
```
gbalabasquer, line 292: I'd first actually define `slice` and then go from there. I feel is a bit arbitrary that `slice` doesn't get defined in the documentation, when `owe` is.

src/dog.sol
```diff
@@ -149,6 +149,73 @@ contract Dog {
...
+    //
+    //     tab <= min(ilk.hole - ilk.dirt, Hole - Dirt)
+    //
+    // The debt of a vault is expressed in normalized terms:
+    //
+    //     vault.debt = vault.art * ilk.rate
+    //
+    // Furthermore, this debt is decreased after a liquidation is triggered:
+    //
+    //     vault.debt' = vault.debt - tab
```
gbalabasquer, line 191: I think there is an error here, `tab` includes the `chop`, that shouldn't count for the remaining debt of the vault.

e18r: Good point, will review

src/dog.sol
```diff
@@ -149,6 +149,77 @@ contract Dog {
...
+    // When a liquidation is successfully started, its `tab` value is added to
+    // the collateral's `dirt` value:
+    //
+    //     ilk.dirt' = ilk.dirt + tab
+    //
+    // To prevent too many liquidations from being triggered at once, the `tab`
+    // value of new liquidations added to the collateral's `dirt` shouldn't
+    // surpass the collateral's `hole` value:
+    //
+    //     tab + ilk.dirt <= ilk.hole
```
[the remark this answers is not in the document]
gbalabasquer: I know this will be taken care below, and it's probably difficult to express the whole thing differently, but if one reads just this equation will be misleading.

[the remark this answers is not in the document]
e18r: I'm handling the dusty case separately, that's why `dust` isn't included there. But I agree it's confusing: it'd be interesting to be able to handle both cases with a single formula, but it doesn't seem to be an easy task.

gbalabasquer:
I think `bark` could be simplified showing the following math blocks. However I'm not 100% sure it accomplishes what we are looking for with all of this.

Main math blocks:

This is quite simple as explaining that whatever normalized debt we take from the vault multiplied by the rate and the liquidation penalty, that will give us the max value to be recovered in an auction.
```
tab = dart * ilk.rate * ilk.chop
```
Remaining debt in the vault will be: 
```
urn.art' = urn.art - dart
```
In this case, the same thing, whatever is the value of `dart`, we are taking the collateral proportional to the debt is being reduced.
```
dink = urn.ink * dart / urn.art
```
Remaining collateral in the vault will be: 
```
urn.ink' = urn.ink - dink
```

We have the room auxiliar value:
```
room = min(Hole - Dirt, ilk.hole - ilk.dust)
```

Then `dart` comes from this (only if `room > 0 && room >= ilk.dust`):

```
dart = urn.art         if urn.art <= room || urn.art > room && urn.art < room + dust
dart = room            otherwise (if urn.art > room && urn.art >= room + dust )
```

Invariants:

```
dink <= urn.ink
```

```
dart <= urn.art
```

```
dart / urn.art == dink / urn.ink == urn.art' / urn.art == urn.ink' / urn.ink
```

```
tab <= min(ilk.hole + ilk.dust, Hole + max(ilk[*].dust)) (considering `ilk[*].dust` can't change after set up)
```

src/dog.sol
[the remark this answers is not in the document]
gbalabasquer: Left a comment in the general dialog to simplify `bark`. Let's continue the discussion there.

src/abaci.sol
```diff
@@ -71,6 +71,20 @@ contract LinearDecrease is Abacus {
...
 
+    
+    
+    
+    // Price calculation when price is decreased linearly in proportion to time:
+    // tau: The number of seconds after the start of the auction where the price will hit 0
+    // top: Initial price
+    // dur: current seconds since the start of the auction
+    
+    // Returns y = top * ((tau - dur) / tau)
```
e18r, line 82: @WilfredTA don't you think that `top * (1 - dur / tau)` is clearer?

src/abaci.sol
```diff
@@ -149,6 +163,14 @@ contract StairstepExponentialDecrease is Abacus {
...
     }
 
+    // top: initial price
+    // dur: seconds since the auction has started
+    // step: seconds between a price drop
+    // cut: percentage to decrease per step (more accurately, the % / 100..e.g., if we want a 2.4% decrease, then cut is 0.024)
+
+    // returns: top * (cut ^ (dur / step))
+
+
```
e18r, line 173: @WilfredTA there are formatting inconsistencies between these two comment blocks: initial description, newlines, capitalization and using `y` as return value.

PR #199, opened 2021-02-03
New formulas for bark and take

gbalabasquer:


src/dog.sol
```diff
@@ -151,88 +151,34 @@ contract Dog {
...
+    // otherwise the transaction fails
+    //
+    // Then `dart` is defined as:
+    // if urn.art <= room || urn.art > room && urn.art < room + dust
+    //      dart = urn.art
+    // otherwise
+    //      dart = room / ilk.rate / ilk.chop
+    //
+    // Leaving the final formulas as:
+    // if urn.art <= room || urn.art > room && urn.art < room + dust:
```
e18r, line 176: `urn.art * ilk.rate` in order for it to be comparable to `room`

gbalabasquer: It should have the `* ilk.chop` as well. Pushed a commit for it.

gbalabasquer:
@e18r what do you say? should we merge this PR in yours? Or still want to check it?

e18r:
OK

PR #200, opened 2021-02-03
feat: added step to dog for LR != CR support

hexonaut:
Not super sold on the name `step`, but basically this small addition will enable to set the liquidation ratio to be less than the vat collateralization ratio. If this has to be pushed after the initial LIQ-2.0 launch so be it, but I figured it was worth discussing to maybe sneak it in while we are still finalizing deployment.

The `step` multiplication is performed to the `art` instead of the `art * rate` so as to not to hit the overflow limit.

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -62,6 +62,7 @@ contract Dog {
     // --- Data ---
     struct Ilk {
         address clip;  // Liquidator
+        uint256 step;  // Liquidation Ratio                                            [wad]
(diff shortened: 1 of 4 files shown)


godsflaw:
We think this is a good change, but it wasn't discussed in `MIP45` and is a heavier change since the code freeze.  There is a change to have an `ilk` based `sin` queue in https://github.com/makerdao/dss/pull/147 that will require us to update the `vow` and the `dog` in the future.  We were thinking we should queue this up for that release.

PR #204, opened 2021-03-03
Don't set unstable default values in stairstep constructor

brianmcmichael:
Addresses QSP-12.2 - The values set in the constructor are not valid parameters. Simplest solution here is to just remove them.


godsflaw (approved):

gbalabasquer (approved):
LGTM

e18r (approved):
LGTM

PR #205, opened 2021-03-03
Require top > 0 for kicking and resetting

gbalabasquer:


src/test/clip.t.sol
```diff
@@ -505,6 +505,22 @@ contract ClipperTest is DSTest {
         assertEq(vat.dai(bob), rad(1000 ether) + rad(100 ether) + tab * 0.02 ether / WAD); // Paid (tip + due * chip) amount of DAI for calling bark()
     }
 
+    function testFail_kick_zero_price() public {
+        pip.poke(bytes32(0));
+        dog.bark(ilk, me, address(this));
```
kmbarry1, line 510: Would be better to test by calling `kick` directly.

gbalabasquer: Yeah, I was a bit lazy on setting up the necessary to `kick` it directly. But I did check the error is coming from that new require.
Btw, trying it directly could also trick us with the auth `require`. So not really sure if worth it.

godsflaw: I'm okay with this so long as the real error was validated this time.

godsflaw (approved):

brianmcmichael (approved):

PR #209, opened 2021-03-05
validation for chop

kmbarry1:
Both Quantstamp and ToB have pointed to potential issues if `chop` is zero or just very small (e.g. liquidations being disabled). There's also a very strong correctness argument that it should never be less than `WAD` due to auction grinding. IMO in this case a little input validation is justified.


brianmcmichael (approved):
I'm ok w/this.

WilfredTA (approved):
lgtm

godsflaw (approved):
I don't like it, but I approve

PR #212, opened 2021-03-10
Add list(), remove getId()

brianmcmichael:
Adds the list() function back in since there's no auto-generated function for it, but uses the auto-generated `active(pos)` function instead of `getId(pos)`


gbalabasquer (approved):
LGTM

kmbarry1 (approved):
LGTM

PR #214, opened 2021-03-10
Gas Optimization: Cache the dust value and allow public modification.

brianmcmichael:
Calling the vat struct to get the dust requires 5 SLOADs and is therefore expensive.

This offloads the `dust` getter to a contract variable that can be publicly modified.

When calling take and redo this will reduce the dust access to a single SLOAD.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -69,6 +69,7 @@ contract Clipper {
     uint256 public cusp;  // Percentage drop before auction reset              [ray]
     uint64  public chip;  // Percentage of tab to suck from vow to incentivize keepers [wad]
     uint192 public tip;   // Flat fee to suck from vow to incentivize keepers          [rad]
+    uint256 public dust;  // Cache the ilk dust to prevent excessive SLOAD
 
     uint256   public kicks;   // Total auctions
     uint256[] public active;  // Array of active auction ids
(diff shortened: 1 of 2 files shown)


godsflaw (commented):
I have no objections to this.  Seems like anyone can update the cache, and we can change the exec lib to call `updust()`.  Short on time, but this looks good so far, want to do one more pass.

kmbarry1 (approved):
LGTM

The gas savings probably outweighs the management overhead. I'm going to consider doing the same sort of thing for `chop` so that we can do the `dust` accounting consistently here as is done in the `Dog`.

gbalabasquer:
> LGTM
> 
> The gas savings probably outweighs the management overhead. I'm going to consider doing the same sort of thing for `chop` so that we can do the `dust` accounting consistently here as is done in the `Dog`.

What do you mean with this?

kmbarry1:
> > LGTM
> > The gas savings probably outweighs the management overhead. I'm going to consider doing the same sort of thing for `chop` so that we can do the `dust` accounting consistently here as is done in the `Dog`.
> 
> What do you mean with this?

I'm referring to the fact that for consistency, we should really be incorporating `chop` into the `dust` check in `take`.

godsflaw (approved):

PR #215, opened 2021-03-12
Return lot size with getStatus

brianmcmichael:
Keeper accessibility improvement. Returns the available `lot` with `getStatus()` so a keeper only needs to make one call to get the price and lot size before performing a `take`

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -442,7 +442,7 @@ contract Clipper {
     }
 
     // Externally returns boolean for if an auction needs a redo and also the current price
-    function getStatus(uint256 id) external view returns (bool needsRedo, uint256 price, uint256 lot) {
+    function getStatus(uint256 id) external view returns (bool needsRedo, uint256 price, uint256 lot, uint256 tab) {
         // Read auction data
         address usr = sales[id].usr;
         uint96  tic = sales[id].tic;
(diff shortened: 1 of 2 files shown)


EdNoepel (approved):
This would prevent keepers from needing to call `sales` along with `status` to determine whether an auction _can_ be bid upon, and if so how much collateral is available to take.

src/clip.sol
```diff
@@ -449,6 +449,7 @@ contract Clipper {
         (done, price) = status(tic, sales[id].top);
 
         needsRedo = usr != address(0) && done;
+        lot = sales[id].lot;
```
kmbarry1, line 452: Would it be interesting to return `tab` as well? (Since that determines the maximum amount of collateral that can be bought.)

Or maybe we should just return `max(lot, tab / price)`?

godsflaw: I think we should just return the `tab` too.

brianmcmichael: This is right, open question whether we want `lot` and `tab` or just the amount of collateral available at that price.

godsflaw: so `function getStatus(uint256 id) external view returns (bool needsRedo, uint256 price, uint256 lot, uint256 tab)`

brianmcmichael: Added tab return

godsflaw (approved):

src/clip.sol
kmbarry1: Yeah, all three of `tab/lot/price` makes the most sense, as then Keepers can just do whatever calculations they want, including the one I described above.

kmbarry1 (approved):
LGTM

PR #216, opened 2021-03-12
Validate ilk when filing clip in Dog

kmbarry1:
Address a ToB finding. Collateral can become permanently stuck in a Clipper if it's for a different `ilk` than the immutable value the Clipper holds.


godsflaw (approved):
LGTM

brianmcmichael (approved):
LGTM

PR #213, opened 2021-03-10
More efficient lock in case tx reverts or if in future there are not …

gbalabasquer:
…refunds

PR #217, opened 2021-03-16
Revert "More efficient lock in case tx reverts or if in future there …

brianmcmichael:
…are not … (#213)"

This reverts commit 628d6f90b4735d99b4c777b6b356b8723c89a67d.


godsflaw (approved):
LGTM on CI 🥗

PR #218, opened 2021-03-16
basic rate-limited hole

kmbarry1:
DO NOT MERGE, FOR DISCUSSION PURPOSES ONLY

Here's a basic implementation of a `hole` that decays in a stepwise linear fashion.

It's stepwise because a continuous decrease would likely lead to too many small auctions in a case where we bumped into a `hole` limit.

Every `qntm` seconds, `dirt` can decrease by up to `clod`. This is triggered in `bark` (adds a lot of `SSTORE` ops, so incentives will matter). But the decrease is only allowed if enough DAI has been collected from auctions or is no longer associated with an auction. This is tracked as `trid` (`dirt` that has been "reversed"), so that if auctions are failing to recapitalize, we don't continue to liquidate. I think there may be an argument for removing this and simplifying to a purely time-based mechanism, but wanted to start from this formulation.

### DIFF
--- a/src/dog.sol
+++ b/src/dog.sol
@@ -66,6 +66,11 @@ contract Dog {
         uint256 chop;  // Liquidation Penalty                                          [wad]
         uint256 hole;  // Max DAI needed to cover debt+fees of active auctions per ilk [rad]
         uint256 dirt;  // Amt DAI needed to cover debt+fees of active auctions per ilk [rad]
+        uint256 trid;  // Amt of dirt dug by auctions.                                 [rad]
+        // Packed in to a single slot to minimze gas costs.
+        uint224 clod;  // Amount of dirt to remove per time quantum                    [rad]
+        uint32  qntm;  // Time between removal of clods                                [seconds]
+        uint256 last;  // Timstamp of most recent dirt reduction                       [Unix epoch]
     }
 
     VatLike immutable public vat;  // CDP Engine
@@ -111,6 +116,10 @@ contract Dog {
 
     function min(uint256 x, uint256 y) internal pure returns (uint256 z) {
         z = x <= y ? x : y;
+    }
+    function min(uint256 x, uint256 y, uint256 z) internal pure returns (uint256 w) {
+        w = x <= y ? x : y;
+        if (z < w) w = z;
     }
     function add(uint256 x, uint256 y) internal pure returns (uint256 z) {
         require((z = x + y) >= x);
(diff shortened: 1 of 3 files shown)


src/dog.sol
```diff
@@ -195,15 +204,36 @@ contract Dog {
...
+                (,rate, spot,, dust) = vat.ilks(ilk);
+                require(spot > 0 && mul(ink, spot) < mul(art, rate), "Dog/not-unsafe");
+            }
+
+            uint256 _Dirt = Dirt;
+            {
+                // Risk of overflow should be extremely small, even if milk.last = 0;
+                // can anyway be made less likely to overflow with a little more work.
+                uint256 pile = mul(sub(block.timestamp, milk.last) / milk.qntm, milk.clod);
+                pile = min(milk.dirt, milk.trid, pile);
```
e18r, line 218: given `trid <= dirt`, why is `dirt` needed here?

src/dog.sol
```diff
@@ -66,6 +66,11 @@ contract Dog {
         uint256 chop;  // Liquidation Penalty                                          [wad]
         uint256 hole;  // Max DAI needed to cover debt+fees of active auctions per ilk [rad]
         uint256 dirt;  // Amt DAI needed to cover debt+fees of active auctions per ilk [rad]
+        uint256 trid;  // Amt of dirt dug by auctions.                                 [rad]
+        // Packed in to a single slot to minimze gas costs.
+        uint224 clod;  // Amount of dirt to remove per time quantum                    [rad]
+        uint32  qntm;  // Time between removal of clods                                [seconds]
```
e18r, line 72: how about removing this and setting `clod` in a per-second basis to save some gas?

e18r:
So the assumption here is that during sharp price declines `trid` will be bigger than `pile`, so liquidations will be rate-limited; but in other circumstances, such as high network congestion, `pile > trid`, so we still want to use the current system and wait for liquidations to finish before allowing more. Am I understanding this correctly?

e18r:
After thinking about this a lot, I think that making `dirt` completely dependent on time is the best option, as you suggest. This would save gas, make things much more predictable, and solve the main issue we're trying to solve.

`digs` would disappear and `dirt` would only be updated by `bark`, after subtracting the decay and adding the `tab` of every new liquidation if it fits in the room.

What I really like about this option, is that it would be a pretty small code change, possibly easier to audit and verify.

PR #220, opened 2021-03-18
Add a breaker level that prevents kick and redo but not take

kmbarry1:
We want this to use in conjunction with the permissionless price cliff breaker being developed in https://github.com/makerdao/clipper-mom/pull/2 , since neither `kick` nor `redo` is safe in a spurious price drop scenario, but allowing already-started auctions to be taken should be fine (since the bad price is not yet in the system when the breaker is triggered).


e18r:
Maybe there's no need for a new level. We could change the level of `redo` from `2` to `1`. Makes this an even more trivial change.

Unless we're envisioning a scenario where there should be no new auctions, but existing ones should be able to be restarted.

kmbarry1:
> Maybe there's no need for a new level. We could change the level of `redo` from `2` to `1`. Makes this an even more trivial change.
> 
> Unless we're envisioning a scenario where there should be no new auctions, but existing ones should be able to be restarted.

I considered that as well; I prefer the flexibility. Turning off `kick` but not `redo` or `take` may be useful in an upgrade pattern.

src/test/clip.t.sol
```diff
@@ -1228,13 +1234,19 @@ contract ClipperTest is DSTest {
...
+        // Any level of stoppage prevents kicking.
         clip.file("stopped", 1);
+        assertTrue(!try_bark(ilk, me));
 
-        dog.bark(ilk, me, address(this));
+        clip.file("stopped", 2);
+        assertTrue(!try_bark(ilk, me));
+
+        clip.file("stopped", 3);
+        assertTrue(!try_bark(ilk, me));
```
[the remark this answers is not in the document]
kmbarry1: That's probably redundant with the other tests, but I suppose it wouldn't hurt.

src/test/clip.t.sol
```diff
@@ -1298,6 +1323,23 @@ contract ClipperTest is DSTest {
...
+        pip.poke(bytes32(uint256(3 ether))); // Spot = $1.50 (update price before reset is called)
+
+        (,,,, uint96 ticBefore, uint256 topBefore) = clip.sales(1);
+        assertEq(uint256(ticBefore), startTime);
+        assertEq(topBefore, ray(5 ether)); // $4 spot + 25% buffer = $5 (wasn't affected by poke)
+
+        hevm.warp(startTime + 3600 seconds);
+        assertTrue(!try_redo(1, address(this)));
+        hevm.warp(startTime + 3601 seconds);
+        assertTrue(try_redo(1, address(this)));
```
gbalabasquer, line 1343: I'm not really sure I'm following these tests, they are `testFail` but we are using the `try` type of functions which would allow us to avoid them. We can better remove the `Fail` and just do `assertTrue(!try_redo,...`
Same for the prev test.

Btw, this kind of test would be even better with a `try` function which actually only tests if gonna succeed without actually keeping the change. In `rwa-example` they were implemented. Basically you need some wrapper that executes the change via a `call`, checks it fails or not, and if it doesn't fail make a `revert`.

src/test/clip.t.sol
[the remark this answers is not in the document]
gbalabasquer: Yes I guessed it was assumed as tested in another test but felt it could be like a good closure for this one.

src/test/clip.t.sol
kmbarry1: Hmmmm yeah something is weird about these, I didn't pay too close attention and just followed the pattern...

kmbarry1: Okay fixed the issue...classic example of why `testFail` is an antipattern except in extremely simple cases.

gbalabasquer (approved):
LGTM

gbalabasquer:
This is gonna failed when merged to `master` due the changes in the `needsRedo` function. Please merge `master` here so the PR can be ready to be merged.

gbalabasquer (changes requested):
Merge from `master` to fix change in needsRedo

kmbarry1:
Rebased on `liq-2.0` and fixed the compilation errors.

gbalabasquer (approved):
LGTM

kmbarry1:
Will wait for another approval before merging.

WilfredTA (approved):
lgtm

godsflaw (approved):
LGTM

brianmcmichael (approved):
My tuppence.

PR #221, opened 2021-03-24
Dog and Clipper comment updates

kmbarry1:
`Dog.bark()`: makes the comment consistent with the code, taking a more high-level, natural language-based approach.
`Clipper.kick()`: improves the comments for a couple of arguments.

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -228,8 +228,8 @@ contract Clipper {
     function kick(
         uint256 tab,  // Debt                   [rad]
         uint256 lot,  // Collateral             [wad]
-        address usr,  // Liquidated CDP
-        address kpr   // Keeper that called dog.bark()
+        address usr,  // Address that will receive any leftover collateral
+        address kpr   // Address that will receive kick/redo incentives
     ) external auth lock isStopped(1) returns (uint256 id) {
         // Input validation
         require(tab  >          0, "Clipper/zero-tab");
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -229,7 +229,7 @@ contract Clipper {
         uint256 tab,  // Debt                   [rad]
         uint256 lot,  // Collateral             [wad]
         address usr,  // Address that will receive any leftover collateral
-        address kpr   // Address that will receive kick/redo incentives
+        address kpr   // Address that will receive incentives
     ) external auth lock isStopped(1) returns (uint256 id) {
         // Input validation
         require(tab  >          0, "Clipper/zero-tab");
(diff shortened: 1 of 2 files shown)


src/dog.sol
```diff
@@ -155,37 +155,18 @@ contract Dog {
...
-    //
-    // Leaving the final formulas as:
-    // if urn.art * ilk.rate * ilk.chop <= room ||
-    //    urn.art * ilk.rate * ilk.chop > room && urn.art * ilk.rate * ilk.chop < room + dust:
-    //      tab = urn.art * ilk.rate * ilk.chop
-    //      lot = urn.ink
-    // otherwise
-    //      tab = room
-    //      lot = urn.ink * (room / ilk.rate / ilk.chop) / urn.art
+    // The entire Vault will be liquidated except when when the DAI target of the
```
e18r, line 162: when ~when~

e18r, line 162: DAI target of => target amount of DAI to be raised in

src/dog.sol
```diff
@@ -155,37 +155,18 @@ contract Dog {
...
-    //      tab = urn.art * ilk.rate * ilk.chop
-    //      lot = urn.ink
-    // otherwise
-    //      tab = room
-    //      lot = urn.ink * (room / ilk.rate / ilk.chop) / urn.art
+    // The entire Vault will be liquidated except when when the DAI target of the 
+    // resulting auction (debt of Vault + liquidation penalty) causes either Dirt to exceed
+    // Hole or ilk.dirt to exceed ilk.hole by an economically significant amount. In that
+    // case, a partial liquidation is performed to respect the limits on outstanding DAI
+    // target globally and per-ilk. The one exception is if the resulting auction would likely
```
e18r, line 166: to respect the limits on outstanding DAI target globally and per-ilk => to respect the global and per-ilk limits on outstanding DAI target

e18r, line 166: ~likely~

kmbarry1: No, "likely" is correct here. There is no guarantee that just because a Vault has more debt than `dust`, that it also has enough collateral to be worth bidding on. It is only a heuristic.

src/clip.sol
```diff
@@ -228,8 +228,8 @@ contract Clipper {
     function kick(
         uint256 tab,  // Debt                   [rad]
         uint256 lot,  // Collateral             [wad]
-        address usr,  // Liquidated CDP
-        address kpr   // Keeper that called dog.bark()
+        address usr,  // Address that will receive any leftover collateral
+        address kpr   // Address that will receive kick/redo incentives
```
gbalabasquer, line 232: actually this is the parameter of `kick`, so it should just be `// Address that will receive incentives` and probably apply the same comment for `redo`

kmbarry1: Yeah, good point. Fixed!

gbalabasquer (approved):
LGTM

e18r:
LGTM

PR #222, opened 2021-03-25
properly account for chop in dust checks in the Clipper

kmbarry1:
We corrected how `dust` was handled in the Dog, but not the Clipper, in the course of the audits. This PR shows what fully making this consistent looks like. It is maybe too late to put it in (as it's IMO a low severity issue, just an inconsistency that could occasionally lead to dusty amounts of collateral left by partial `takes`, which can happen anyway e.g. if a dusty Vault is liquidated), but I thought everyone should at least understand why the current code is inconsistent. If we launch with that inconsistency, it can potentially be corrected in Liq 2.1, which also gives us more time to refactor into a more gas-efficient format (a good argument against this change is that the small inconsistency is acceptable in order to reduce gas costs).

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -365,11 +365,12 @@ contract Clipper {
                 slice = owe / price;        // slice' = owe' / price <= owe / price == slice <= lot
             } else if (owe < tab && slice < lot) {
                 // If slice == lot => auction completed => dust doesn't matter
-                if (tab - owe < chost) {     // safe as owe < tab
+                uint256 chost_ = chost;
+                if (tab - owe < chost_) {    // safe as owe < tab
                     // If tab <= chost, buyers have to take the entire lot.
-                    require(tab > chost, "Clipper/no-partial-purchase");
+                    require(tab > chost_, "Clipper/no-partial-purchase");
                     // Adjust amount to pay
-                    owe = tab - chost;       // owe' <= owe
+                    owe = tab - chost_;      // owe' <= owe
                     // Adjust slice
                     slice = owe / price;     // slice' = owe' / price < owe / price == slice < lot
                 }
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -451,10 +451,11 @@ contract Clipper {
         done  = (sub(block.timestamp, tic) > tail || rdiv(price, top) < cusp);
     }
 
-    // Public function to update the cached dust value.
+    // Public function to update the cached dust*chop value.
     function upchost() external {
-        (,,,, uint256 _dust) = VatLike(vat).ilks(ilk);
-        chost = wmul(_dust, dog.chop(ilk));
+        _ilk = ilk;  // cache
+        (,,,, uint256 _dust) = VatLike(vat).ilks(_ilk);
+        chost = wmul(_dust, dog.chop(_ilk));
     }
 
     // Cancel an auction during ES or via governance action.
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -309,7 +309,7 @@ contract Clipper {
     // Buy up to `amt` of collateral from the auction indexed by `id`.
     // 
     // Auctions will not collect more DAI than their assigned DAI target,`tab`;
-    // thus, if `amt` would cost more DAI than `tab` at the current price,, the
+    // thus, if `amt` would cost more DAI than `tab` at the current price, the
     // amount of collateral purchased will instead be just enough to collect `tab` DAI.
     //
     // To avoid partial purchases resulting in very small leftover auctions that will
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -442,7 +442,7 @@ contract Clipper {
     }
 
     // Externally returns boolean for if an auction needs a redo and also the current price
-    function getStatus(uint256 id) external view returns (bool needsRedo, uint256 price) {
+    function getStatus(uint256 id) external view returns (bool needsRedo, uint256 price, uint256 lot, uint256 tab) {
         // Read auction data
         address usr = sales[id].usr;
         uint96  tic = sales[id].tic;
@@ -451,6 +451,8 @@ contract Clipper {
         (done, price) = status(tic, sales[id].top);
 
         needsRedo = usr != address(0) && done;
+        lot = sales[id].lot;
+        tab = sales[id].tab;
     }
 
     // Internally returns boolean for if an auction needs a redo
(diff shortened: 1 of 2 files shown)


src/clip.sol
```diff
@@ -294,7 +295,7 @@ contract Clipper {
         uint256 coin;
         if (_tip > 0 || _chip > 0) {
             uint256 _dust = dust;
-            if (tab >= _dust && mul(lot, price) >= _dust) {
```
gbalabasquer, line 297: I think we should define `chost` here as well and use for both the `tab` and `lot` to make it even more consistent.

gbalabasquer:
I see now what you mean, and it makes sense to change it.

src/clip.sol
kmbarry1: Well, it would be conceptually wrong for the `lot` check...although that one is not right as it is...it should use `mat` to adjust higher the `dust` value, since a Vault with `dust` amount of debt should have around `mat * dust` DAI worth of collateral at liquidation.

kmbarry1: Tbh I'm tempted to say the check here on `tab` is redundant, and we should just have the `lot` check (with the `mat` correction), since it is "amount of collateral that's worth bidding on" that `dust` is supposed to be a proxy for anyway.

gbalabasquer: Well the concept of dusty `lot` could be whatever we want. We thought to use the exact same value that would make an auction dusty from the DAI perspective. So if the debt to be recovered is less than X amount is considered dusty, then the same value should be considered for the `lot` at the current price. That's why I consider it should be used the same `chust` value.
About `mat` I don't know, seems a bit complicated and at that point the relationship between the remaining collateral and debt could be anything. It could have been purchased a part of the collateral at a much lower price than the liquidation price (a part from the liquidation penalty).

gbalabasquer: > Tbh I'm tempted to say the check here on `tab` is redundant, and we should just have the `lot` check (with the `mat` correction), since it is "amount of collateral that's worth bidding on" that `dust` is supposed to be a proxy for anyway.

I'll think a bit more about this tomorrow, but I think the opposite might happen as well. It would be a very rare case though and it would require governance increasing the `dust` in the middle.
I agree the most important check is the `lot` one.

kmbarry1: I'd argue that although we can define "dusty `lot`" however we want, we should do so based on the most sensible criterion. And in my opinion, that's "value of collateral is too small to be interesting to bid on, even if buying all of it". But that's exactly what the `dust` value is intended to correspond to. But, `dust` is a debt amount--i.e. if you set the ETH-A `dust` to 100 DAI, you're saying that the marginal Vault (i.e. one that gets liquidated when its debt is exactly 100 DAI) should have an amount of collateral that is interesting to bid on, economically speaking. For ETH-A, that means 150 DAI worth of ETH due to the collateralization ratio of 150%. So I think the right comparison for whether `price * lot` is dusty would be `rmul(dust, mat)`.

kmbarry1: Okay, thinking about it more, actually it depends on exactly how `dust` is being set...it might be set in way such that the auction of a near-dusty Vault still returns collateral to the user, minus penalty. And in that case comparing against `chost` as you suggest actually is correct.

gbalabasquer: So my reasoning is the following:

Let's say for example we have a `dust` of 10,000 and 13% `chop..
In general we those parameters we could agree that any CDP that would for leave an auction with less of 11,300 `tab` wouldn't be very interesting.
So when we analyze how much can be left in the auction on partial `take`, we should use the same 11,300 DAI value instead of 10,000 DAI. That's basically the fix you are doing here in the `take` function.

On the other side, some time ago, we decided that in order to pay the incentive, the remaining auction needs to keep being interesting, meaning that the value of purchasing collateral has to be at least 11,300 DAI as well (there your other fix).
IMO the collateral should be treated the same way. It needs to be enough collateral to at least be able to collect that amount of DAI at the "market" price (we use the `osm` to get it). At that moment I don't think we care about the `mat` itself at all.
We just have a remaining `tab` and `lot`, we want both of them to be above a certain threshold, and to be consistent that should be `mul(price, chost)` (11,130 DAI worth of collateral).

kmbarry1: I see where you are coming from, but think about how `dust` is chosen: https://forum.makerdao.com/t/signal-request-adjusting-dust-parameter-2021-2/6704

The goal is to guarantee that there's some minimum _amount of collateral_ present in each auction, which makes bidding on a marginally dusty Vault attractive up to some maximum gas price and collateral price drop post-liquidation. Look in particular how a higher `dust` value is recommended for ETH-B in the forum post above due to its smaller `mat`.

Of course, as you point out, by the time `redo` is being called, the proportion of `tab` in relation to the collateral may have changed significantly, so applying the exact same logic as used at liquidation isn't really sound.

What we we're really trying to accomplish is: only give rewards for calling `redo` if the Keepers will actually bid on the reset auction. Whether we're comparing the collateral value at current price to `dust`, `wmul(dust, chop)`, or `rmul(dust, mat)`, it's all just a heuristic, because the true value is dynamic and depends on gas prices, Keepers' cost of capital, current actual market price of collateral and DAI, etc. So, for now maybe we just pick something simple and go with it. This honestly might need to be its own risk parameter, which we could introduce in Liq 2.1.

gbalabasquer: I agree this might require its own value but as you said for 2.0 we should keep it simple.
I think using the same exact one than for the `tab` would be the simpler to reason about at this stage. I mean, between comparing to `dust * price` or `chost * price` I'd prefer the later.

kmbarry1: Yeah I think we're in agreement there. Also, if we use `chost` everywhere, we can store `chost` instead of `dust`, and change `updust` to `upchost`. The downside is we have to update the value whenever either `dust` or `chop` changes, and it will costs a bit more gas as well, but that should still be rare, so probably worth it to save an `SLOAD` in `take`.

Do you think we should make this change before launch?

gbalabasquer: It good be nice to keep consistent this value everywhere.

kmbarry1:
I've fleshed this out into a fully-tested form that can be considered for merge.

e18r (approved):
LGTM

gbalabasquer (approved):
LGTM

brianmcmichael (approved):

PR #224, opened 2021-03-30
rename getPrice() to getFeedPrice() and price to feedPrice

e18r:
Note that this could be inconsistent with daiwanese practices.

Check out also the [daiwanese version](https://github.com/makerdao/dss/pull/225)

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -210,11 +210,11 @@ contract Clipper {
     // Could get this from rmul(Vat.ilks(ilk).spot, Spotter.mat()) instead, but
     // if mat has changed since the last poke, the resulting value will be
     // incorrect.
-    function getFeedPrice() internal returns (uint256 price) {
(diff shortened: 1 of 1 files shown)


gbalabasquer (approved):
LGTM

src/clip.sol
```diff
@@ -210,7 +210,7 @@ contract Clipper {
     // Could get this from rmul(Vat.ilks(ilk).spot, Spotter.mat()) instead, but
     // if mat has changed since the last poke, the resulting value will be
     // incorrect.
-    function getPrice() internal returns (uint256 price) {
+    function getFeedPrice() internal returns (uint256 price) {
```
kmbarry1, line 213: Might as well name the return var `feedPrice`.

e18r: done!

kmbarry1 (approved):
I prefer a meme name but not gonna block on that.

PR #225, opened 2021-03-30
rename getPrice() to feed() and price to feed

e18r:
Note I'm not fluent in daiwanese.

Check out also the [non-daiwanese version](https://github.com/makerdao/dss/pull/224)

### DIFF
--- a/src/clip.sol
+++ b/src/clip.sol
@@ -210,11 +210,11 @@ contract Clipper {
     // Could get this from rmul(Vat.ilks(ilk).spot, Spotter.mat()) instead, but
     // if mat has changed since the last poke, the resulting value will be
     // incorrect.
-    function getPrice() internal returns (uint256 price) {
+    function feed() internal returns (uint256 feed) {
         (PipLike pip, ) = spotter.ilks(ilk);
         (bytes32 val, bool has) = pip.peek();
         require(has, "Clipper/invalid-price");
-        price = rdiv(mul(uint256(val), BLN), spotter.par());
+        feed = rdiv(mul(uint256(val), BLN), spotter.par());
     }
 
     // start an auction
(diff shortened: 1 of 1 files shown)


src/clip.sol
```diff
@@ -210,11 +210,11 @@ contract Clipper {
     // Could get this from rmul(Vat.ilks(ilk).spot, Spotter.mat()) instead, but
     // if mat has changed since the last poke, the resulting value will be
     // incorrect.
-    function getPrice() internal returns (uint256 price) {
+    function feed() internal returns (uint256 feed) {
```
gbalabasquer, line 213: There is a shadowed variable warning here. I'd go with the other PR which is much clearer what is the value.

kmbarry1, line 213: `feed` seems like it would denote an oracle itself, not a price. If we're doing meme naming, I'd invent a new meme here, ...maybe `mark`, derived from "market price". cc @gbalabasquer

gbalabasquer: I gave my approval to the other PR. At this point I'd prefer we focus in the other PRs than thinking on this variable naming. The other PR looks good to me.

kmbarry1: I mean, the shadowing is easy to resolve...just don't use a declared return var with the same name as the function.

gbalabasquer: Yeah, but also not sure if you can do `uint256 feed = feed();`.

gbalabasquer: If it's the case you could solve it with `uint256 feed_ = feed();` though

gbalabasquer: Whatever you prefer, if we find good daiwanese names for distinguishing both type of prices I'm ok. Just don't want to spend much time on it. I'm not sure if `mark` says much to me. IMO in this type of cases where we have two completely different values but with a very similar meaning is difficult to use daiwanese.

e18r:
closing in favor of the non-daiwanese version

PR #230, opened 2021-04-30
cmooney-20210430: Add note about skim()/thaw()

godsflaw:
This issue came up when running the cage-keeper in `kovan`.  The documentation needs to be more clear.

### DIFF
--- a/src/end.sol
+++ b/src/end.sol
@@ -148,6 +148,11 @@ interface SpotLike {
     all auctions are in the reverse (`dent`) phase. There are two ways
     of ensuring this:
 
+    NOTE: In the event there's a system surplus, and there are no
+    under-collateralised vaults remaining, one must skim() large CDPs in order
+    to account for the fees built up in the surplus buffer.  Without doing this
+    the thaw() will fail to complete as it requires the surplus buffer be empty.
+
     4a. i) `wait`: set the cooldown period to be at least as long as the
            longest auction duration, which needs to be determined by the
            cage administrator.


EdNoepel (changes requested):
Please adjust line 135 to clarify that not only undercollateralized  urns should be skimmed.

src/end.sol
```diff
@@ -148,6 +148,11 @@ interface SpotLike {
     all auctions are in the reverse (`dent`) phase. There are two ways
     of ensuring this:
 
+    NOTE: In the event there's a system surplus, and there are no
+    under-collateralised vaults remaining, one must skim() large CDPs in order
```
EdNoepel, line 152: Is it only _large_ CDPs?  Seems like if we need 0 surplus we need to `skim` all non-empty CDPs.

godsflaw: We shouldn't need to `skim()` everything, as we just need enough to collect fees get surplus to 0.  The surplus cap means some of those fees are owed, but were sent off to surplus auctions.  That is, `surplus < all fees owed`.

src/end.sol
```diff
@@ -148,6 +148,11 @@ interface SpotLike {
     all auctions are in the reverse (`dent`) phase. There are two ways
     of ensuring this:
```
EdNoepel, line 149: This line references two options on old line 151 and 158.  The new comment seems to be wedged in the middle

gbalabasquer: Yeah, I'd suggest moving the new paragraph right after the `- backing collateral taken`

PR #235, opened 2021-09-16
Move to GH Actions

dizzy:
output of passed build https://github.com/makerdao/dss/runs/3624027429


.travis.yml
```diff
@@ -1,10 +0,0 @@
-os:
-  - linux
-language: nix
-nix: 2.3.6
```
gbalabasquer, line 4: I remember there was a reason to pin the version 2.3.6. Not sure if still necessary but would be a problem to put it fixed just in case?

dizzy: yep, that's pinned now by using `install_url: https://releases.nixos.org/nix/nix-2.3.6/install`

cristidas (approved):

PR #236, opened 2021-09-21
Add events to join adapters

brianmcmichael:
ds-note was removed from `master`, this adds the necessary events to the join adapters.

### DIFF
--- a/src/join.sol
+++ b/src/join.sol
@@ -66,8 +66,14 @@ interface VatLike {
 contract GemJoin {
     // --- Auth ---
     mapping (address => uint) public wards;
-    function rely(address usr) external auth { wards[usr] = 1; }
-    function deny(address usr) external auth { wards[usr] = 0; }
+    function rely(address usr) external auth {
+        wards[usr] = 1;
+        emit Rely(usr);
+    }
+    function deny(address usr) external auth {
+        wards[usr] = 0;
+        emit Deny(usr);
+    }
     modifier auth {
         require(wards[msg.sender] == 1, "GemJoin/not-authorized");
         _;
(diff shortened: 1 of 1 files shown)


src/join.sol
```diff
@@ -79,35 +85,52 @@ contract GemJoin {
     uint    public dec;
     uint    public live;  // Active Flag
 
+    // Events
+    event Rely(address indexed usr);
+    event Deny(address indexed usr);
+    event Join(address indexed urn, uint256 wad);
```
talbaneth, line 91: just wondering - are we sure we want the event parameter (`urn`) to have a different name than the function parameter (`usr`)?

src/join.sol
```diff
@@ -79,35 +85,52 @@ contract GemJoin {
...
+    event Exit(address indexed guy, uint256 wad);
```
talbaneth, line 92: Also here we log `guy` while exit() gets `usr`.

talbaneth (commented):
Looks good, see question about difference between the events and functions parameters (just to make sure we made an aware decision).

src/join.sol
talbaneth: I'm thinking this is probably ok, just acknowledge you are aware and I'll resolve and approve.

gbalabasquer: This should be `usr`.

src/join.sol
gbalabasquer: This should be `usr`.

src/join.sol
```diff
@@ -117,6 +140,13 @@ contract DaiJoin {
     DSTokenLike public dai;  // Stablecoin Token
     uint    public live;     // Active Flag
 
+    // Events
+    event Rely(address indexed usr);
+    event Deny(address indexed usr);
+    event Join(address indexed urn, uint256 wad);
+    event Exit(address indexed guy, uint256 wad);
```
gbalabasquer, line 147: Same here. I know this might be a bit confusing wit the `usr` of `Rely` and `Deny` but I think I still prefer to keep the same naming than the functions.

brianmcmichael: Yeah, this is a bigger issue where the different join adapters use inconsistent variable names. I was going to go through in a separate PR and clean all of this up but I'll just do it here.

talbaneth (approved):
LGTM

src/join.sol
brianmcmichael: Everything updated to `usr` here.

gbalabasquer (approved):
LGTM

PR #238, opened 2021-11-02
Freeze dai déployment addresse

brianmcmichael:
Edit:
I pushed to this PR by @brianmcmichael some test changes that will freeze the déployment addresse of dai in hevm tests. This compléments nicely the original intent of this PR.

Original:
Adds script used in signature generation for testing. This is currently available [here](https://github.com/MrChico/tokens/blob/master/scripts/generateSig.js) but it gets lost some times and we use it to update the signatures in test when hevm updates. Putting it here so that it doesn't get lost.


gbalabasquer:
I've pushed a fix for the CI.
So using the original address as fixed one, doesn't it make sense to keep the script here?
I mean we shouldn't need to upgrade the addresses anymore. Maybe we could just push the scripts to a makerdao repo with a more representative name just in case we want to access it.

julienmartinlevrai:
Thank you for fixing the CI !
As for moving the script away, I agree it doesn’t need to be here, but maybe we can leave it here for now until we find a more appropriate répo

julienmartinlevrai:
What do you think, @brianmcmichael ?

brianmcmichael:
It's useful as documentation for calculating permit signatures for other purposes. There's very little documentation on this generally.

brianmcmichael:
I can't approve the changes here though since I opened the PR.

julienmartinlevrai:
@gbalabasquer if you agree let us leave it here for now while we figure out a better place to put it

PR #243, opened 2021-12-27
Change vow to split out accounting from auction logic

hexonaut:
Similar to https://github.com/makerdao/dss/pull/241 this is designed to make the `vow` immutable going forward, but instead of using a proxy it splits things out into two contracts. Read the other PR for the reason why this is important. First step is to get this working same as before. After that we can fix things like the `kiss(...)` debt blocking bug.

WIP Tests are still failing and code is not complete.

Notes:

 * Compared to the proxy vow this will require updating the liquidations to use the `bow` instead of the `vow`.
 * Also, naming maybe `sow` is better than `bow` ? No reason for `bow` in particular it was just the first thing that came into my head.

### DIFF
--- a/src/bow.sol
+++ b/src/bow.sol
@@ -1,0 +1,170 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+/// bow.sol -- Dai settlement module
+
+// Copyright (C) 2018 Rain <rainbreak@riseup.net>
+// Copyright (C) 2021-2022 Dai Foundation
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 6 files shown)


talbaneth:
Small note - dog.sol will need to change to use both `vow` and `bow`.
Also any UI/bot that integrates with flap (and flop) auctions will still rely on the `bow` address and will have to update each time it changes.

rontinchie:
they commit mistakes on my contract: now they trying you install and hack my account, they need to resolve this mistake in proper way they need to compensate my wallet addresses when they commit mistakes, if public will know this , crypto might be lost forever as they will the trust to cryto currency and when the government find this one out im sure all the big names of exchanges and wallet will suffer from a tougher regulations. all of my friends kmow this even my family so in case there something happen to me, one of them will spit this anomally.

salvotoshi77-creator (approved):

PR #244, opened 2022-01-24
Rate Limited Flap

hexonaut:
This will prevent auctions from being kicked off if the amount in auction (`fill`) has gone above the `lid`.

### DIFF
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -69,7 +69,7 @@ contract Flapper is LibNote {
     uint48   public   tau = 2 days;   // 2 days total auction length  [seconds]
     uint256  public kicks = 0;
     uint256  public live;  // Active Flag
-    uint256  public limit; // max dai to be in auction at one time  [rad]
+    uint256  public lid;   // max dai to be in auction at one time  [rad]
     uint256  public usage; // current dai in auction                [rad]
 
     // --- Events ---
@@ -106,7 +106,7 @@ contract Flapper is LibNote {
         if (what == "beg") beg = data;
         else if (what == "ttl") ttl = uint48(data);
         else if (what == "tau") tau = uint48(data);
-        else if (what == "limit") limit = data;
+        else if (what == "lid") lid = data;
         else revert("Flapper/file-unrecognized-param");
     }
 
@@ -115,7 +115,7 @@ contract Flapper is LibNote {
         require(live == 1, "Flapper/not-live");
         require(kicks < uint(-1), "Flapper/overflow");
         usage = add256(usage, lot);
-        require(usage <= limit, "Flapper/over-limit");
+        require(usage <= lid, "Flapper/over-lid");
         id = ++kicks;
 
         bids[id].bid = bid;
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -3,6 +3,7 @@
 /// flap.sol -- Surplus auction
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
+// Copyright (C) 2022 Dai Foundation
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -69,6 +69,8 @@ contract Flapper is LibNote {
     uint48   public   tau = 2 days;   // 2 days total auction length  [seconds]
     uint256  public kicks = 0;
     uint256  public live;  // Active Flag
+    uint256  public limit; // max dai to be in auction at one time  [rad]
+    uint256  public usage; // current dai in auction                [rad]
 
     // --- Events ---
     event Kick(
@@ -89,6 +91,12 @@ contract Flapper is LibNote {
     function add(uint48 x, uint48 y) internal pure returns (uint48 z) {
         require((z = x + y) >= x);
     }
+    function add256(uint x, uint y) internal pure returns (uint z) {
+        require((z = x + y) >= x);
+    }
+    function sub(uint x, uint y) internal pure returns (uint z) {
+        require((z = x - y) <= x);
+    }
     function mul(uint x, uint y) internal pure returns (uint z) {
         require(y == 0 || (z = x * y) / y == x);
     }
@@ -98,6 +106,7 @@ contract Flapper is LibNote {
         if (what == "beg") beg = data;
         else if (what == "ttl") ttl = uint48(data);
         else if (what == "tau") tau = uint48(data);
+        else if (what == "limit") limit = data;
         else revert("Flapper/file-unrecognized-param");
     }
 
@@ -105,6 +114,8 @@ contract Flapper is LibNote {
     function kick(uint lot, uint bid) external auth returns (uint id) {
         require(live == 1, "Flapper/not-live");
         require(kicks < uint(-1), "Flapper/overflow");
+        usage = add256(usage, lot);
+        require(usage <= limit, "Flapper/over-limit");
         id = ++kicks;
 
         bids[id].bid = bid;
@@ -143,9 +154,11 @@ contract Flapper is LibNote {
     function deal(uint id) external note {
         require(live == 1, "Flapper/not-live");
         require(bids[id].tic != 0 && (bids[id].tic < now || bids[id].end < now), "Flapper/not-finished");
-        vat.move(address(this), bids[id].guy, bids[id].lot);
+        uint256 lot = bids[id].lot;
+        vat.move(address(this), bids[id].guy, lot);
         gem.burn(address(this), bids[id].bid);
         delete bids[id];
+        usage = sub(usage, lot);
     }
 
     function cage(uint rad) external note auth {
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -158,7 +158,11 @@ contract Flapper is LibNote {
         vat.move(address(this), bids[id].guy, lot);
         gem.burn(address(this), bids[id].bid);
         delete bids[id];
-        usage = sub(usage, lot);
+        if (usage >= lot) {
+            usage = sub(usage, lot);
+        } else {
+            usage = 0;
+        }
     }
 
     function cage(uint rad) external note auth {
--- a/src/flap.sol
+++ b/src/flap.sol
@@ -70,7 +70,7 @@ contract Flapper is LibNote {
     uint256  public kicks = 0;
     uint256  public live;  // Active Flag
     uint256  public lid;   // max dai to be in auction at one time  [rad]
-    uint256  public usage; // current dai in auction                [rad]
+    uint256  public fill;  // current dai in auction                [rad]
 
     // --- Events ---
     event Kick(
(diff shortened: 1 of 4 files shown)


src/flap.sol
```diff
@@ -98,13 +106,16 @@ contract Flapper is LibNote {
         if (what == "beg") beg = data;
         else if (what == "ttl") ttl = uint48(data);
         else if (what == "tau") tau = uint48(data);
+        else if (what == "limit") limit = data;
```
brianmcmichael, line 109: what about `lid` for proper daiwanese

src/flap.sol
```diff
@@ -143,9 +154,11 @@ contract Flapper is LibNote {
     function deal(uint id) external note {
         require(live == 1, "Flapper/not-live");
         require(bids[id].tic != 0 && (bids[id].tic < now || bids[id].end < now), "Flapper/not-finished");
-        vat.move(address(this), bids[id].guy, bids[id].lot);
+        uint256 lot = bids[id].lot;
+        vat.move(address(this), bids[id].guy, lot);
         gem.burn(address(this), bids[id].bid);
         delete bids[id];
+        usage = sub(usage, lot);
```
brianmcmichael, line 161: Something for the team to think about is if we can ever get into a situation where `usage` is lower than `lot`. I don't see anything immediately but we will freeze up auctions if this gets mismatched.

src/flap.sol
hexonaut: Sure, I don't have strong preference for names.

src/flap.sol
```diff
@@ -89,6 +91,12 @@ contract Flapper is LibNote {
     function add(uint48 x, uint48 y) internal pure returns (uint48 z) {
         require((z = x + y) >= x);
     }
+    function add256(uint x, uint y) internal pure returns (uint z) {
```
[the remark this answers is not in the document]
brianmcmichael: I'd push back but it sounds like we may be replacing this with a clapper anyways so ok.

src/flap.sol
brianmcmichael: Yeah trying to think if it's worthwhile adding a `file` for usage but I don't think it is because people will be tempted to use it and that will really mess things up. In the unexpected case this does happen we can just eat the loss on the particular action and deploy a new flap.

hexonaut: I added the conditional.

src/test/flap.t.sol
```diff
@@ -150,4 +153,34 @@ contract FlapTest is DSTest {
...
+        // Just up to the lid
+        flap.kick({ lot: 500 ether
+                  , bid: 0
+                  });
+        // Just over the lid
+        flap.kick({ lot: 1
+                  , bid: 0
+                  });
+    }
+    function test_deal() public {
```
brianmcmichael, line 171: Can you add another test very similar to this that has more than one auction running at once and that asserts the intermediate state?

talbaneth:
Just a note to remember - we will need to leave the old flapper authorized on the vat so ongoing flaps can finish.
Can probably deny it as the following spell.

src/flap.sol
```diff
@@ -69,6 +69,8 @@ contract Flapper is LibNote {
     uint48   public   tau = 2 days;   // 2 days total auction length  [seconds]
     uint256  public kicks = 0;
     uint256  public live;  // Active Flag
+    uint256  public lid;   // max dai to be in auction at one time  [rad]
+    uint256  public usage; // current dai in auction                [rad]
```
iamchrissmith, line 73: daiwanese attempt: `usage` -> `fill`?

iamchrissmith (commented):
need to update the copyright to include Dai Foundations 2022

src/flap.sol
hexonaut: Actually, this conditional feels weird and a waste of gas. I think I'll leave it out.

talbaneth:
> Just a note to remember - we will need to leave the old flapper authorized on the vat so ongoing flaps can finish. Can probably deny it in the following spell.

on second thought don't think that's needed, it doesn't do anything pemissive on the vat.

src/flap.sol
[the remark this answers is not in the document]
hexonaut: The overload doesn't work for some reason. Try it yourself.

src/test/flap.t.sol
hexonaut: This is done.

src/flap.sol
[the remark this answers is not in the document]
iamchrissmith: weird, I see that now too.  It seems to be able to figure out the 256 one but not the 48 one... I wonder if it has to do with argument packing or something where at that level uint48 looks like a uint256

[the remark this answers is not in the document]
gbalabasquer: I think the problem is that you can pass a `uint48` to a function that expects a `uint48` but also a  higher `uint`. So the code doesn't know what to use in that case.

gbalabasquer:
> > Just a note to remember - we will need to leave the old flapper authorized on the vat so ongoing flaps can finish. Can probably deny it in the following spell.
> 
> on second thought don't think that's needed, it doesn't do anything permissive on the vat.

> Just a note to remember - we will need to leave the old flapper authorized on the vat so ongoing flaps can finish. Can probably deny it in the following spell.

Right, the only vat authorization is a `hope` from the `vow`, that one will be revoked at the moment the `flapper` is replaced. But that shouldn't be a problem as only needed in `kick`.

gbalabasquer (changes requested):
There is one test failing in `vat.t.sol` and one in `vow.t.sol` regarding flaps.

iamchrissmith:
so as I was looking at this this morning I was a little worried about governance changing the `vow.bump` (flap `lot` size) and how that might mess up our adding to and subtracting from the `fill`.  I think I convinced myself that the math would work out there even if governance changes the `bump` since we are always adding before subtracting and since we are reading from `bid.lot`. 

However, I wanted to raise the possibility of having a different measurement, what if `lid` and `fill` were in number of auctions and not amount of dai.  We would never need to worry about `bump`/`lot` impacts, I think it would save us some variables and might make reasoning a little easier.  It seems governance/risk would be making the same calculation (how much dai/maker can the market support at once).  I am not wedded to it, but wanted to raise the idea to see how others felt.

gbalabasquer:
> so as I was looking at this this morning I was a little worried about governance changing the `vow.bump` (flap `lot` size) and how that might mess up our adding to and subtracting from the `fill`. I think I convinced myself that the math would work out there even if governance changes the `bump` since we are always adding before subtracting and since we are reading from `bid.lot`.
> 
> However, I wanted to raise the possibility of having a different measurement, what if `lid` and `fill` were in number of auctions and not amount of dai. We would never need to worry about `bump`/`lot` impacts, I think it would save us some variables and might make reasoning a little easier. It seems governance/risk would be making the same calculation (how much dai/maker can the market support at once). I am not wedded to it, but wanted to raise the idea to see how others felt.

Before finding those flap tests errors I was going to say that I expected this to be a counter instead of a max sum but this way might be a bit more flexible for governance. I think it looks good as it is, but yeah I initially pictured it as a simple auctions counter.

talbaneth:
> so as I was looking at this this morning I was a little worried about governance changing the `vow.bump` (flap `lot` size) and how that might mess up our adding to and subtracting from the `fill`. I think I convinced myself that the math would work out there even if governance changes the `bump` since we are always adding before subtracting and since we are reading from `bid.lot`.
> 
> However, I wanted to raise the possibility of having a different measurement, what if `lid` and `fill` were in number of auctions and not amount of dai. We would never need to worry about `bump`/`lot` impacts, I think it would save us some variables and might make reasoning a little easier. It seems governance/risk would be making the same calculation (how much dai/maker can the market support at once). I am not wedded to it, but wanted to raise the idea to see how others felt.

bwt I also convinced myself yesteday about the `lot` substraction being safe (from the same reasons you mentioned)

hexonaut:
Fixed the tests.

iamchrissmith:
> > so as I was looking at this this morning I was a little worried about governance changing the `vow.bump` (flap `lot` size) and how that might mess up our adding to and subtracting from the `fill`. I think I convinced myself that the math would work out there even if governance changes the `bump` since we are always adding before subtracting and since we are reading from `bid.lot`.
> > However, I wanted to raise the possibility of having a different measurement, what if `lid` and `fill` were in number of auctions and not amount of dai. We would never need to worry about `bump`/`lot` impacts, I think it would save us some variables and might make reasoning a little easier. It seems governance/risk would be making the same calculation (how much dai/maker can the market support at once). I am not wedded to it, but wanted to raise the idea to see how others felt.
> 
> Before finding those flap tests errors I was going to say that I expected this to be a counter instead of a max sum but this way might be a bit more flexible for governance. I think it looks good as it is, but yeah I initially pictured it as a simple auctions counter.

That is a good point gonza! I think the flexibility of using a counter for governance could be important.  It could be messy if governance wanted to change the `bump` and the `lid` at the same time because we could end up where we have a difference in the `fill` and `lid` that remains until old flaps clear... Maybe not the biggest issue, but I think I am starting to feel like I can see small reasons to use a counter and I'm not sure I see a benefit of the sum.

talbaneth:
Per the discussion at the end of office hours, we should add a `live` variable that can block functionality (at least `fire()` and `denyProxy()`). This will make sure that when we switch the ESM nothing wrong can happen even if we don't deny a contract that the old ESM was relied on.

gbalabasquer (approved):
LGTM

iamchrissmith:
> > > > so as I was looking at this this morning I was a little worried about governance changing the `vow.bump` (flap `lot` size) and how that might mess up our adding to and subtracting from the `fill`. I think I convinced myself that the math would work out there even if governance changes the `bump` since we are always adding before subtracting and since we are reading from `bid.lot`.
> > > > However, I wanted to raise the possibility of having a different measurement, what if `lid` and `fill` were in number of auctions and not amount of dai. We would never need to worry about `bump`/`lot` impacts, I think it would save us some variables and might make reasoning a little easier. It seems governance/risk would be making the same calculation (how much dai/maker can the market support at once). I am not wedded to it, but wanted to raise the idea to see how others felt.
> > > 
> > > 
> > > Before finding those flap tests errors I was going to say that I expected this to be a counter instead of a max sum but this way might be a bit more flexible for governance. I think it looks good as it is, but yeah I initially pictured it as a simple auctions counter.
> > 
> > 
> > That is a good point gonza! I think the flexibility of using a counter for governance could be important. It could be messy if governance wanted to change the `bump` and the `lid` at the same time because we could end up where we have a difference in the `fill` and `lid` that remains until old flaps clear... Maybe not the biggest issue, but I think I am starting to feel like I can see small reasons to use a counter and I'm not sure I see a benefit of the sum. I created a PR for this change to see how people feel about that version of the code: #245
> 
> Sorry Chris, I think I expressed my self wrongly. Wanted to say that this way is probably more flexible than the counter. But honestly is the same exact thing. If governance wants to change the `bump`, they can do it in tandem with `lid`. It will depend on the case if it would be easier to use a counter or an accumulator.

No, rereading it, that is what you said, I just misread it, sorry

brianmcmichael:
One more test at https://github.com/makerdao/dss/pull/246

PR #245, opened 2022-01-25
change lid to a count of auctions

iamchrissmith:
Change to a counter instead of dai measure for the `lid`


hexonaut:
Well we are focused on the amount of MKR required, and not so much the number of auctions which is why I chose DAI amount. We want to keep the limit such that the regular automated keepers have enough capital on hand to bid on all the auctions.

Changing the `bump` on the `vow` shouldn't be a problem as every `kick()` has a matching `deal()` of the same amount even if the `bump` changes part way through.

PR #246, opened 2022-01-31
add test of lid alteration during auction

brianmcmichael:
Adds a test of reducing the `lid` while the auction is active to assert we don't get into a stuck state.

### DIFF
--- a/src/test/flap.t.sol
+++ b/src/test/flap.t.sol
@@ -208,4 +208,24 @@ contract FlapTest is DSTest {
         flap.deal(id2);
         assertEq(flap.fill(), 300 ether);
     }
+    function test_mod_lid_in_flight() public {
+        uint256 id = flap.kick({ lot: 400 ether
+                  , bid: 0
+                  });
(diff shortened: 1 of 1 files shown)


src/test/flap.t.sol
```diff
@@ -208,4 +208,24 @@ contract FlapTest is DSTest {
...
+        uint256 id = flap.kick({ lot: 400 ether
+                  , bid: 0
+                  });
+        assertEq(flap.fill(), 400 ether);
+
+        // Reduce lid while auction is active
+        flap.file("lid", 300 ether);
+
+        Guy(ali).tend(id, 400 ether, 1 ether);
+        assertEq(flap.fill(), 300 ether);
```
gbalabasquer, line 221: This should still be 400 ether

brianmcmichael: fixed

gbalabasquer (approved):
LGTM

iamchrissmith (approved):
lgtm

hexonaut (approved):

PR #247, opened 2022-01-31
Add Cure

gbalabasquer:


src/cure.sol
```diff
@@ -0,0 +1,125 @@
...
+        live = 0;
+        emit Cage();
+    }
+
+    function reset(address src) external {
+        uint256 amt = data[src].amt;
+        if (amt > 0) {
+            total = _sub(total, amt);
+        }
+        total = _add(total, data[src].amt = _toUint128(SourceLike(src).cure()));
```
iamchrissmith, line 123: for consistency with the `addSource` pattern, should we do:

```suggestion
        data[src].amt = _toUint128(SourceLike(src).cure());
        if (data[src].amt > 0) {
             total = _add(total, data[src].amt);   
        } 
```

gbalabasquer: Hmm I think in this case we would be doing two extra calls to SLOAD (which are cheaper now but still more expensive than not doing them). So what we can do is:
```suggestion
    data[src].amt = amt = _toUint128(SourceLike(src).cure());
    if (amt > 0) {
        total = _add(total, amt);   
    }
```

iamchrissmith: oh yeah that works

talbaneth:
I'm wondering if we want to time the Cure.reset() calls to be done only at a specific time in the ES process.
This is imprtant since we don't want to use non updated data from the sources.
Also unless I'm missing something there's no use for this module before the ES process.
For that I think we should assume that the data in the sources will not change after the ES waiting period, so it's fine to read at that point. 

To implement this we can add to End a new function before thaw() to call into Cure and have a flag in Cure that allows reset only after that (this also means we will have to auth End on Cure)
That new function in End will be the one checking the waiting period and blocking thaw() with a new flag in End.

gbalabasquer:
> I'm wondering if we want to time the Cure.reset() calls to be done only at a specific time in the ES process. This is imprtant since we don't want to use non updated data from the sources. Also unless I'm missing something there's no use for this module before the ES process. For that I think we should assume that the data in the sources will not change after the ES waiting period, so it's fine to read at that point.
> 
> To implement this we can add to End a new function before thaw() to call into Cure and have a flag in Cure that allows reset only after that (this also means we will have to auth End on Cure). That new function in End will be the one checking the waiting period and blocking thaw() with a new flag in End.
> 
> at thaw() stage Cure will need to then check that all sources have been updated. Can keep a counter for that, and allow each source to be updated only once (thus simplifying the code a bit).

Well it is certainly another approach which might have some advantages with the trade-off of complicating a bit the module and it is relationship with `End` (which is super simple now).
However a part from those extra complexities we would be introducing again the dependency that if any of the `reset` calls fail due a failure in `module.cure()` the whole thing will be stuck. I know we could put a `try/catch` but starts to feel a bit over-engineered.  Anyway let's see what others think about it.

src/cure.sol
```diff
@@ -0,0 +1,128 @@
...
+        emit Deny(usr);
+    }
+
+    function addSource(address src) external auth isLive {
+        Source storage data_ = data[src];
+        require(data_.pos == 0, "Cure/already-existing-source");
+        sources.push(src);
+        data_.pos = _toUint128(sources.length);
+        data_.amt = _toUint128(SourceLike(src).cure());
+        if (data_.amt > 0) {
```
talbaneth, line 92: Might be controversial but for very rarely called functions (which is all logic in this contract :)) I would remove the gas optimization of checking if X > 0 each time before updating total.
Will make the code more compact, and I think that's also the general approach taken in other core contracts.

talbaneth:
> > I'm wondering if we want to time the Cure.reset() calls to be done only at a specific time in the ES process. This is imprtant since we don't want to use non updated data from the sources. Also unless I'm missing something there's no use for this module before the ES process. For that I think we should assume that the data in the sources will not change after the ES waiting period, so it's fine to read at that point.
> > To implement this we can add to End a new function before thaw() to call into Cure and have a flag in Cure that allows reset only after that (this also means we will have to auth End on Cure). That new function in End will be the one checking the waiting period and blocking thaw() with a new flag in End.
> > at thaw() stage Cure will need to then check that all sources have been updated. Can keep a counter for that, and allow each source to be updated only once (thus simplifying the code a bit).
> 
> Well it is certainly another approach which might have some advantages with the trade-off of complicating a bit the module and it is relationship with `End` (which is super simple now). However a part from those extra complexities we would be introducing again the dependency that if any of the `reset` calls fail due a failure in `module.cure()` the whole thing will be stuck. I know we could put a `try/catch` but starts to feel a bit over-engineered. Anyway let's see what others think about it.

Certainly agree about preventing any chance this will block ES.
As discussed one scenario where timing is critical is making sure we sample/reset only after L2 ES hanshake protocol is completed (where I think the waiting period is assumed to be larger than the L2 slow path, but not sure by how much).

src/cure.sol
```diff
@@ -0,0 +1,128 @@
...
+        require((y = uint128(x)) == x, "Cure/toUint128-overflow");
+    }
+
+    constructor() public {
+        live = 1;
+        wards[msg.sender] = 1;
+        emit Rely(msg.sender);
+    }
+
+    function sLength() external view returns (uint256 size) {
```
talbaneth, line 72: Neat: consider changing `Length` to `Count`.
(Length tends to refer to contiguous elements while Count tends to refer to a number of elements)

src/cure.sol
```diff
@@ -0,0 +1,128 @@
...
+        size = sources.length;
+    }
+
```
talbaneth, line 75: Consider addint a getter for the list (as in https://github.com/makerdao/dss/blob/master/src/clip.sol#L435).

src/cure.sol
```diff
@@ -0,0 +1,128 @@
...
+contract Cure {
+    mapping (address => uint256) public wards;
+    uint256 public live;
+    address[] public sources;
+    uint256 public total;
+    mapping (address => Source) public data;
+
+    struct Source {
+        uint128 pos;
```
talbaneth, line 34: To try and avoid confusion maybe add a comment such as:
"// position in sources plus 1, 0 means a source does not exist"

SImilarly to:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol#L43

src/cure.sol
```diff
@@ -0,0 +1,128 @@
...
+        uint128 amt;
```
talbaneth, line 35: mm this is a RAD (at least treated as such in END), so cannot be store as uint128.

talbaneth: To avoid the use of uint128 we can use a separate positions array.
Will be more gas costly but I don't think it matters so much specifically in this contract.

gbalabasquer: Done.

src/cure.sol
gbalabasquer: Done.

src/cure.sol
gbalabasquer: Done.

src/cure.sol
gbalabasquer: Yes you are right, it is doesn't make much sense here. Done.

src/cure.sol
```diff
@@ -0,0 +1,115 @@
...
+    function deny(address usr) external auth isLive {
+        wards[usr] = 0;
+        emit Deny(usr);
+    }
+
+    function addSource(address src) external auth isLive {
+        require(pos[src] == 0, "Cure/already-existing-source");
+        sources.push(src);
+        pos[src] = sources.length;
+        uint256 amt_ = amt[src] = SourceLike(src).cure();
```
talbaneth, line 86: Either way is probably ok, but I would remove the cure sampling here, thus separating the logic of adding  a source and sampling it. Just seems to me a bit cleaner.
If we want to sample (/reset) as part of the spell we can just add that call right after.

talbaneth:
Changes look good.
Added another suggestion and a test request.

src/cure.sol
gbalabasquer: Yes you have a good point, however I'd prefer to leave this one open for now to see others' opinions before applying the change.

talbaneth:
Looks good to me pending the discussions of timing reset calls and calling reset upon adding a source.

src/cure.sol
hexonaut: Yeah I think I'm leaning more towards leaving it out to separate the logic. `reset()` will only need to be called after shutdown really.

src/cure.sol
```diff
@@ -0,0 +1,115 @@
...
+            sources[pos_ - 1] = move;
+            pos[move] = pos_;
+        }
+        sources.pop();
+        total = _sub(total, amt[src]);
+        delete pos[src];
+        delete amt[src];
+    }
+
+    function cage() external auth isLive {
```
hexonaut, line 105: I'm assuming this is here for ESM, and this will be authed on the ESM module for governance attack scenario. Can you confirm that is the only reason?

hexonaut (changes requested):
Some comments on design as well as one critical bug.

I think a non-zero `cure` test should be added to the `end` tests as well.

src/cure.sol
```diff
@@ -0,0 +1,115 @@
...
+        live = 0;
+        emit Cage();
+    }
+
+    function reset(address src) external {
```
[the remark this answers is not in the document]
gbalabasquer: Oops great catch Sam, can't believe I missed this one.

src/cure.sol
```diff
@@ -0,0 +1,115 @@
...
+pragma solidity >=0.6.12;
+
+interface SourceLike {
+    function cure() external view returns (uint256);
+}
+
+contract Cure {
+    mapping (address => uint256) public wards;
+    uint256 public live;
+    address[] public sources;
```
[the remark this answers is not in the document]
gbalabasquer: Yeah let's keep it custom at least in this case.

src/cure.sol
gbalabasquer: Yes that's the idea. The other option is to make it authed in the `End` module and call `cure.cage()` in `end.cage()` directly.

src/cure.sol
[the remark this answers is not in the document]
gbalabasquer: Fixed now. I was so focused on the functions to add or remove a source that missed the most important one.

gbalabasquer:
> Some comments on design as well as one critical bug.
> 
> I think a non-zero `cure` test should be added to the `end` tests as well.

Yes planning to add some cure ones on the `end` tests as soon as we finish modeling it.

talbaneth:
@hexonaut @kmbarry1 and others, would be good to hear opinions about whether we want to time the `Cure.reset()` calls to be done only at a specific time in the ES process (and probably rename `Cure.reset()` to something else). See discussion above.
I guess if we do opt for that then it's a big change and better be done sooner rather than later.

src/cure.sol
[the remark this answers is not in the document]
kmbarry1: Looks like the check was implemented as `pos[src] == 1`, but shouldn't it be `pos[src] > 0`? Otherwise `reset` can only ever be called for the first source.

src/cure.sol
kmbarry1: I like the idea of calling it in `End.cage` because 1) the End already has to store its address, and 2) this is a single global module. Then no Keeper / non-malicious governance actor needs to worry about caging the Cure separately.

kmbarry1: Although I see some conflicting comments suggesting that maybe we want the Cure to be callable after shutdown in some scenarios--maybe I will answer this for myself as I progress through the review, but if anyone wants to give me the quick rundown that might save time.

src/cure.sol
kmbarry1: I don't have terribly strong feelings here, it's almost just a matter of aesthetics. Might personally lean towards a call to `reset` in `addSource` to avoid duplicating logic but quickly confirming that the added source supports the correct API. But really I think any option will work.

src/cure.sol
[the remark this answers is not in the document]
kmbarry1: +1 to @gbalabasquer here, let's not profane `dss` with OZ code (or any other external dependency for that matter).

src/end.sol
```diff
@@ -412,7 +418,7 @@ contract End {
         require(debt == 0, "End/debt-not-zero");
         require(vat.dai(address(vow)) == 0, "End/surplus-not-zero");
         require(block.timestamp >= add(when, wait), "End/wait-not-finished");
-        debt = vat.debt();
+        debt = sub(vat.debt(), cure.total());
```
kmbarry1, line 421: Maybe a good defensive change to the logic would be:
```
debt = vat.debt() >= cure.total() ? sub(vat.debt(), cure.total()) : vat.debt();
```
?

I am a little conflicted on it b/c while it protects us against certain accidents (e.g. setting the `cure` for some source to an anomalously huge amount), there are still accidents that will break things, using `vat.debt()` as the fallback is unlikely to be correct, and it doesn't really protect from malicious actors (they can still e.g. make `total == debt`). So after going to the trouble of suggesting it maybe I've talked myself out of it already, but going to post anyway for others to consider (maybe it inspires some more useful safeguard e.g. in the Cure itself).

src/test/cure.t.sol
```diff
@@ -0,0 +1,212 @@
...
+    function testFailRely() public {
+        cure.deny(address(this));
+        cure.rely(address(123));
+    }
+
+    function testFailDeny() public {
+        cure.deny(address(this));
+        cure.deny(address(123));
+    }
+
```
kmbarry1, line 44:
```suggestion
```

src/cure.sol
kmbarry1: Ah I see we intend to allow `reset` only after `cage`. I believe I am still in favor of my earliest comment here advocating to `cage` it in the End.

kmbarry1 (changes requested):
First pass complete; see comments. At least one small thing to fix (in `reset`).

kmbarry1:
> @hexonaut @kmbarry1 and others, would be good to hear opinions about whether we want to time the `Cure.reset()` calls to be done only at a specific time in the ES process (and probably rename `Cure.reset()` to something else). See discussion above. I guess if we do opt for that then it's a big change and better be done sooner rather than later.

Hmmmm I can see a point to both sides. I think that if we require sources to report during settlement, then we must wrap the calls to `cure()` in try-catch statements so a reverting source cannot block ES. As to whether we _should_ require a post-shutdown report from each source... If `reset` can only be called once per source after shutdown, someone might call it too early for one or more sources, resulting in inaccuracies. So that leads me to think the logic should be "`reset` must be called _at least_ once per source after shutdown" to allow for multiple calls if necessary. Tentatively I think that would be okay (if paired with reversion protection), and I think would make it more likely that the ultimate accounting was correct. Note that a simpler way to implement this (i.e. minimizing changes to the End) is to do it entirely in DssCure by making `total` a `private` variable, and making the public fetching function revert after `cage` unless `reset` has been called at least once for all sources (e.g. by keeping a counter similar to Tal's original suggestion, which is only updated by `reset` post-`cage`). What does everyone else think?

kmbarry1:
Maybe the biggest business logic question here is: _should_ it be possible for a single source to block settlement until it achieves some consistency point? I tend to favor "no" (and this is consistent with how the End relies only on a waiting period at the moment for auctions to finish), but if we think the answer is actually "yes", there are a number of compromises and implementation options we can work through (e.g. make it configurable which sources are strictly required to successfully return a value post-`cage`).

talbaneth:
> > @hexonaut @kmbarry1 and others, would be good to hear opinions about whether we want to time the `Cure.reset()` calls to be done only at a specific time in the ES process (and probably rename `Cure.reset()` to something else). See discussion above. I guess if we do opt for that then it's a big change and better be done sooner rather than later.
> 
> Hmmmm I can see a point to both sides. I think that if we require sources to report during settlement, then we must wrap the calls to `cure()` in try-catch statements so a reverting source cannot block ES. As to whether we _should_ require a post-shutdown report from each source... If `reset` can only be called once per source after shutdown, someone might call it too early for one or more sources, resulting in inaccuracies. So that leads me to think the logic should be "`reset` must be called _at least_ once per source after shutdown" to allow for multiple calls if necessary. Tentatively I think that would be okay (if paired with reversion protection), and I think would make it more likely that the ultimate accounting was correct. Note that a simpler way to implement this (i.e. minimizing changes to the End) is to do it entirely in DssCure by making `total` a `private` variable, and making the public fetching function revert after `cage` unless `reset` has been called at least once for all sources (e.g. by keeping a counter similar to Tal's original suggestion, which is only updated by `reset` post-`cage`). What does everyone else think?

I like this (making `total` private).
Note thoug that the requirement for each source to report after `cage()` might not be strict enough since some of the modules will only start their reporting/calculating process after they are caged. We might want to require each source to report after `End`'s waiting period. That can be achieved either by storing `End` in `Cure` and reading `End.wait`, or even simpler by just passing `End.wait` as a parameter to `Cure.cage()`. However we obtain the waiting duration in `Cure`, it will start it's reporting counter only after the time has passed. If we do it this way then since both `Cure.reset` and `End.thaw` are permissionless there is really no point in allowing each source to be reset more than once.

src/cure.sol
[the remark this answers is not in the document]
hexonaut: Just to be clear I suggest copying the code instead of adding some external dependancy such as [here](https://github.com/makerdao/dss-wormhole/blob/master/src/utils/EnumerableSet.sol). In this case I can appreciate the argument against. I'm fine with custom code. Will review it carefully.

src/cure.sol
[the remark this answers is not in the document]
gbalabasquer: Seriously considering taking vacation... (:man_facepalming:). Yes the check should be the same one than in `delSource` and now is fixed.

gbalabasquer:
Hmmm I don't know, I feel we should not over-engineer it (at least for a first version).
The `End` module already heavily depends on keepers. If keepers do not skim underwater CDPs the whole accounting will be screwed. I think this should be implemented in a similar way.
But if you guys feel too strongly about making sure that at least `reset` (or maybe it should just be call `set`) is called at least once per source after `cage` then I'll proceed implementing it.

kmbarry1:
> We might want to require each source to report after `End`'s waiting period.

It sounds like we're still uncertain about the requirements here. To deal with that uncertainty, I'd suggest a configurable delay in `DssCure` (as opposed to locking to `End.wait`), unless we're willing to accept needing to upgrade the Cure in the future. Upgrades shouldn't be too painful unless we have a lot of sources, which won't be the case at first, so probably it's fine either way.

kmbarry1:
Copying over some discussion from the closed PR (in order, Kurt then Gonza then Tal):

> > > I'd break the open questions down as follows:
> > > 
> > > 1. Must all sources be successfully set **at least** once after shutdown?
> > > 2. Must all sources be successfully set **at most** once after shutdown?
> > > 3. Should the Cure attempt to prevent a source from being set "too soon" after shutdown, or should that logic live in the source itself?
> > > 4. Is it acceptable for global settlement to be blocked or delayed by a source that has not been set yet?
> > 
> > 
> > 
> > 1. In theory yes, as otherwise the accounting will be probably broken (excepting the missing source correspond to report 0 value). However it is better to give the DAI holders less than corresponds than giving nothing because the module is stuck due a broken source.
> > 2. I don't see the need to have that constraint.
> > 3. I think it can live in each source. The source might revert on `cure()` if not ready for final number.
> > 4. I think I wouldn't do it. GS needs to happen and not be totally stuck as it can happen if a `src` is not returning for any reason. Maybe what it can be done is there is a max timeout where if all the sources didn't report, the module still reports whatever is calculated as `total`.
> 
> Agree with almost everything. I like the max timeout idea as it removes the need for a try-catch.
> 
> Just regarding (3) - we can have that revert logic in the sources but it seems a bit like duplication (each module should have it) and source for bugs (each such source writer should be aware to do it and cannot just rely on its `cure` storage variable being constantly updated). If we have the max timeout than we store the cage timestamp anyway. So adding another delay storage parameter and checking against it in `set` seems like a small price to pay for being able to control everything from `Cure`.

I think the discussions boil down to two options:
1) Go with the current simple implementation and rely on Keepers doing what they need to do and `wait` being set long enough to allow sources to settle.
2) Implement an "at least once" requirement for sources to report post-shutdown along with a separate waiting period after which lack of reporting no longer blocks reading `total`.

(1) is simpler but (2) gives us a bit more flexibility (can let `wait` be shorter than the max reporting time for Cure sources). I kind of think (1) is probably sufficient as the incentives for Keepers to actually take all required actions isn't that much different between the two. Unless we think it's very important to be able to speed shutdown up in the case that all sources report well before their deadline.

talbaneth:
> Copying over some discussion from the closed PR (in order, Kurt then Gonza then Tal):
> 
> > > > I'd break the open questions down as follows:
> > > > 
> > > > 1. Must all sources be successfully set **at least** once after shutdown?
> > > > 2. Must all sources be successfully set **at most** once after shutdown?
> > > > 3. Should the Cure attempt to prevent a source from being set "too soon" after shutdown, or should that logic live in the source itself?
> > > > 4. Is it acceptable for global settlement to be blocked or delayed by a source that has not been set yet?
> > > 
> > > 
> > > 
> > > 1. In theory yes, as otherwise the accounting will be probably broken (excepting the missing source correspond to report 0 value). However it is better to give the DAI holders less than corresponds than giving nothing because the module is stuck due a broken source.
> > > 2. I don't see the need to have that constraint.
> > > 3. I think it can live in each source. The source might revert on `cure()` if not ready for final number.
> > > 4. I think I wouldn't do it. GS needs to happen and not be totally stuck as it can happen if a `src` is not returning for any reason. Maybe what it can be done is there is a max timeout where if all the sources didn't report, the module still reports whatever is calculated as `total`.
> > 
> > 
> > Agree with almost everything. I like the max timeout idea as it removes the need for a try-catch.
> > Just regarding (3) - we can have that revert logic in the sources but it seems a bit like duplication (each module should have it) and source for bugs (each such source writer should be aware to do it and cannot just rely on its `cure` storage variable being constantly updated). If we have the max timeout than we store the cage timestamp anyway. So adding another delay storage parameter and checking against it in `set` seems like a small price to pay for being able to control everything from `Cure`.
> 
> I think the discussions boil down to two options:
> 
> 1. Go with the current simple implementation and rely on Keepers doing what they need to do and `wait` being set long enough to allow sources to settle.
> 2. Implement an "at least once" requirement for sources to report post-shutdown along with a separate waiting period after which lack of reporting no longer blocks reading `total`.
> 
> (1) is simpler but (2) gives us a bit more flexibility (can let `wait` be shorter than the max reporting time for Cure sources). I kind of think (1) is probably sufficient as the incentives for Keepers to actually take all required actions isn't that much different between the two. Unless we think it's very important to be able to speed shutdown up in the case that all sources report well before their deadline.

If you think we can count on keepers to sample the source (without any incentive and in ES chaos) between the point that  L2 finished reporting their shutdown procedure (say 7 days after cage because of an L2's dispute period) and the End's waiting period (say we push that to 9 days after cage instead of the current 3 days) than ok, we can go with the simpler implementation (1).

kmbarry1:
> If you think we can count on keepers to sample the source (without any incentive and in ES chaos) between the point that L2 finished reporting their shutdown procedure (say 7 days after cage because of an L2's dispute period) and the End's waiting period (say we push that to 9 days after cage instead of the current 3 days) than ok, we can go with the simpler implementation (1).

I think about it like this: the main incentive to make sure sources report is for those who plan to redeem DAI, because under-reporting raises the value of `debt` recorded and thus lowers the amount of collateral disbursed per DAI. I don't see why an extra delay strengthens or weakens that incentive by much--if no one wants to make sources report, they're probably don't mind waiting a couple more days (if the extra delay would be much longer, weeks or months, then it probably does incentivize a speed up). If we think there's a high chance Keepers will just be unaware that sources need to be reported, then that's a strong argument for having "at least once" enforcement + an extra delay, just so they don't forget.

Personally I do think a "sources must report at least once, but after some delay the requirement is voided" logic makes sense from a correctness perspective.

gbalabasquer:
@kmbarry1 @talbaneth please let me know what you think about this last [commit](https://github.com/makerdao/dss/pull/247/commits/2b23cd9fd6a2a61178fce445ead4fdad1ea06cf1)

talbaneth:
I think this is good (did only a logical review of the recent change, did not yet go into varaible names and synthax etc..).

Again, personally I would have made it even stricter by storing an additional wait period which only after that you can `load()`.
But not a must I guess and we can probably add that in the future if we want (a benefit of the pull model is that update is simple).

gbalabasquer:
> I think this is good (did only a logical review of the recent change, did not yet go into varaible names and synthax etc..).
> 
> Again, personally I would have made it even stricter by storing an additional wait period which only after that you can `load()`. But not a must I guess and we can probably add that in the future if we want (a benefit of the pull model is that update is simple).

Yeah, let's try to not complicate more. I'll wait for your final review so we can aim to pass it to CS on Wednesday.

kmbarry1 (dismissed):
LGTM

I think this is good for a first version.

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+    uint256 public wait;
```
talbaneth, line 30: Name is a bit confusing since it's not End.wait, maybe change to `stretch`?

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+    uint256 public when;
```
talbaneth, line 31: Maybe change to `loose`? (just to have something with more meaning)

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+    mapping (address => uint256) public pos; // position in sources + 1, 0 means a source does not exist
+    mapping (address => uint256) public amt;
+    mapping (address => uint256) public loaded;
+    uint256 public loadedNum;
```
talbaneth, line 35: Rename to lcount to conform with `count()` below and use just one word?

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+    uint256 amount_;
```
talbaneth, line 36: Rename to `report_`? (see suggested change of amount() to report() below)

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    constructor() public {
+        live = 1;
+        wards[msg.sender] = 1;
+        emit Rely(msg.sender);
+    }
+
+    function count() external view returns (uint256 count_) {
+        count_ = sources.length;
```
talbaneth, line 65: // There is a small discrepancy between this function and `amount()` where you return without a variable.

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    function list() external view returns (address[] memory) {
+        return sources;
+    }
+
+    function amount() external view returns (uint256) {
```
talbaneth, line 72: Rename to `report()` to remove confusion with amount and amt?

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+        else revert("Cure/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    function addSource(address src) external auth {
+        require(live == 1, "Cure/not-live");
+        require(pos[src] == 0, "Cure/already-existing-source");
+        sources.push(src);
+        pos[src] = sources.length;
+    }
```
talbaneth, line 101: Emit an event?

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+        uint256 last = sources.length;
+        if (pos_ < last) {
+            address move = sources[last - 1];
+            sources[pos_ - 1] = move;
+            pos[move] = pos_;
+        }
+        sources.pop();
+        delete pos[src];
+        delete amt[src];
+    }
```
talbaneth, line 116: Emit an event?

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+        require(live == 0, "Cure/still-live");
+        require(pos[src] > 0, "Cure/non-existing-source");
+        uint256 oldAmt_ = amt[src];
+        uint256 newAmt_ = amt[src] = SourceLike(src).cure();
+        amount_ = _add(_sub(amount_, oldAmt_), newAmt_);
+        if (loaded[src] == 0) {
+            loaded[src] = 1;
+            loadedNum ++;
+        }
+    }
```
talbaneth, line 135: Emit an event?

src/cure.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    function load(address src) external {
+        require(live == 0, "Cure/still-live");
+        require(pos[src] > 0, "Cure/non-existing-source");
+        uint256 oldAmt_ = amt[src];
+        uint256 newAmt_ = amt[src] = SourceLike(src).cure();
+        amount_ = _add(_sub(amount_, oldAmt_), newAmt_);
+        if (loaded[src] == 0) {
+            loaded[src] = 1;
+            loadedNum ++;
```
talbaneth, line 133: Remove space before the `++`?

talbaneth (commented):
See some naming suggestions (subjective, fill free to ignore).
Also some minor synthax stuff.
Finally wondering whether we need events for the sources stuff.

src/cure.sol
gbalabasquer: Done

src/cure.sol
gbalabasquer: Done.

src/cure.sol
gbalabasquer: Done.

src/cure.sol
gbalabasquer: Done.

src/cure.sol
gbalabasquer: Well it's `Cure.wait`. It wouldn't be the first time we use the same variable name for different contracts.

src/cure.sol
gbalabasquer: Same than `wait`.

src/cure.sol
gbalabasquer: Maybe just `loaded`?

gbalabasquer: LOL yeah... Ok let me think if I can think on some better approach otherwise will follow your suggestion.

gbalabasquer: I've renamed `loadedNum` to `lCount` (as you suggested) and `count` to `tCount` (total count).
If we figure better namings, we can always change it later.

talbaneth (dismissed):
LGTM

hexonaut (commented):
Two minor points. Overall looks good.

src/cure.sol
```diff
@@ -27,25 +27,28 @@ contract Cure {
...
     uint256 public live;
     address[] public sources;
-    uint256 public total;
+    uint256 public wait;
+    uint256 public when;
     mapping (address => uint256) public pos; // position in sources + 1, 0 means a source does not exist
     mapping (address => uint256) public amt;
+    mapping (address => uint256) public loaded;
+    uint256 public lCount;
+    uint256 report_;
```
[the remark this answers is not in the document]
hexonaut: `peek` ? I hate doing the naming lol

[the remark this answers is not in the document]
naszam: what about `buzz`, `snap`, `howl` or `copy` ?
ref. [`report` synonyms](https://www.merriam-webster.com/thesaurus/report)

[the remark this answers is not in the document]
gbalabasquer: I've done some renaming and exposing the variable as requested.

talbaneth (approved):
Recent changes LGTM

hexonaut (approved):
LGTM

PR #260, opened 2022-03-02
Set only after cage, and force cage on End.cage

gbalabasquer:
This PR makes the `cure` implementation closer to what you guys have been discussing. Do you think this way is a bit better?
It would be missing the control that every source is set after `cage` (which I'm still not sure if really needed)

### DIFF
--- a/src/cure.sol
+++ b/src/cure.sol
@@ -41,11 +41,6 @@ contract Cure {
         _;
     }
 
-    modifier isLive {
-        require(live == 1, "Cure/not-live");
-        _;
-    }
-
     // --- Internal ---
     function _add(uint256 x, uint256 y) internal pure returns (uint256 z) {
         require((z = x + y) >= x, "Cure/add-overflow");
@@ -69,25 +64,27 @@ contract Cure {
         return sources;
     }
 
-    function rely(address usr) external auth isLive {
+    function rely(address usr) external auth {
+        require(live == 1, "Cure/not-live");
         wards[usr] = 1;
         emit Rely(usr);
     }
 
-    function deny(address usr) external auth isLive {
+    function deny(address usr) external auth {
+        require(live == 1, "Cure/not-live");
         wards[usr] = 0;
         emit Deny(usr);
     }
 
-    function addSource(address src) external auth isLive {
+    function addSource(address src) external auth {
+        require(live == 1, "Cure/not-live");
         require(pos[src] == 0, "Cure/already-existing-source");
         sources.push(src);
         pos[src] = sources.length;
-        uint256 amt_ = amt[src] = SourceLike(src).cure();
-        total = _add(total, amt_);
     }
 
-    function delSource(address src) external auth isLive {
+    function delSource(address src) external auth {
+        require(live == 1, "Cure/not-live");
         uint256 pos_ = pos[src];
         require(pos_ > 0, "Cure/non-existing-source");
         uint256 last = sources.length;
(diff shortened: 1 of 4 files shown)


kmbarry1:
I definitely like the idea to `cage` the Cure from the End. As for only allowing setting source amounts after `cage`--well I don't see any problem with it and it does simplify the code a bit. I am still uncertain whether we need a hard time delay or otherwise a requirement to "only set once" source amounts. I guess we can also take into account the way the Cure is implemented when implementing sources.

gbalabasquer:
> I definitely like the idea to `cage` the Cure from the End. As for only allowing setting source amounts after `cage`--well I don't see any problem with it and it does simplify the code a bit. I am still uncertain whether we need a hard time delay or otherwise a requirement to "only set once" source amounts. I guess we can also take into account the way the Cure is implemented when implementing sources.

Actually I think if we are doing a "only set once" then we need to set a delay, otherwise they could be done too early.
other option is not putting the delay and allow multiple sets per source (as it is now).
The other question is if we want to make sure that at least `set` was called once, that would require an extra storage field.

kmbarry1:
I'd break the open questions down as follows:
1) Must all sources be successfully set **at least** once after shutdown?
2) Must all sources be successfully set **at most** once after shutdown?
3) Should the Cure attempt to prevent a source from being set "too soon" after shutdown, or should that logic live in the source itself?
4) Is it acceptable for global settlement to be blocked or delayed by a source that has not been set yet?

gbalabasquer:
> I'd break the open questions down as follows:
> 
> 1. Must all sources be successfully set **at least** once after shutdown?
> 2. Must all sources be successfully set **at most** once after shutdown?
> 3. Should the Cure attempt to prevent a source from being set "too soon" after shutdown, or should that logic live in the source itself?
> 4. Is it acceptable for global settlement to be blocked or delayed by a source that has not been set yet?

1. In theory yes, as otherwise the accounting will be probably broken (excepting the missing source correspond to report 0 value). However it is better to give the DAI holders less than corresponds than giving nothing because the module is stuck due a broken source.

2. I don't see the need to have that constraint.

3. I think it can live in each source. The source might revert on `cure()` if not ready for final number.

4. I think I wouldn't do it. GS needs to happen and not be totally stuck as it can happen if a `src` is not returning for any reason. Maybe what it can be done is there is a max timeout where if all the sources didn't report, the module still reports whatever is calculated as `total`.

talbaneth:
> I'd break the open questions down as follows:
> 
> 1. Must all sources be successfully set **at least** once after shutdown?
> 2. Must all sources be successfully set **at most** once after shutdown?
> 3. Should the Cure attempt to prevent a source from being set "too soon" after shutdown, or should that logic live in the source itself?
> 4. Is it acceptable for global settlement to be blocked or delayed by a source that has not been set yet?

talbaneth:
> > I'd break the open questions down as follows:
> > 
> > 1. Must all sources be successfully set **at least** once after shutdown?
> > 2. Must all sources be successfully set **at most** once after shutdown?
> > 3. Should the Cure attempt to prevent a source from being set "too soon" after shutdown, or should that logic live in the source itself?
> > 4. Is it acceptable for global settlement to be blocked or delayed by a source that has not been set yet?
> 
> 1. In theory yes, as otherwise the accounting will be probably broken (excepting the missing source correspond to report 0 value). However it is better to give the DAI holders less than corresponds than giving nothing because the module is stuck due a broken source.
> 2. I don't see the need to have that constraint.
> 3. I think it can live in each source. The source might revert on `cure()` if not ready for final number.
> 4. I think I wouldn't do it. GS needs to happen and not be totally stuck as it can happen if a `src` is not returning for any reason. Maybe what it can be done is there is a max timeout where if all the sources didn't report, the module still reports whatever is calculated as `total`.

Agree with almost everything.
I like the max timeout idea as it removes the need for a try-catch.

Just regarding (3) -  we can have that revert logic in the sources but it seems a bit like duplication (each module should have it) and source for bugs (each such source writer should be aware to do it and cannot just rely on its `cure` storage variable being constantly updated).
If we have the max timeout than we store the cage timestamp anyway. So adding another delay storage parameter and checking against it in `set` seems like a small price to pay for being able to control everything from `Cure`.

gbalabasquer:
I'm merging this branch in the main one as seems to be a consensus this is better than previous one. Let's keep the discussions of pending functionality there.

PR #262, opened 2022-03-21
Bugfix drip() in Pot

qzirak:
Small bug fix in pot.sol in the drip() function where the current time can never be equal to rho.


gbalabasquer:
Hmm this is not a bug but an intended feature. `join` requires that the `drip` function is called in the same block prev to it.
Also you mention there is a fix in `drip` but the modification was in `join`.

PR #263, opened 2022-03-21
RPC tests

julienmartinlevrai:


gbalabasquer:
Hmmm what do these changes have to do with the forked integration tests we want to do?

kmbarry1:
+1 to Gonza's points, we want a thing that forks mainnet and runs various actions against that fork. The current contents of `dss` shouldn't be relevant to that--all you need to know is the interfaces to the contracts and their addresses.

derek-flossman:
Create a new repo as a sub-module dependency and pull the new off-chain tests into it, e.g. reference: https://github.com/makerdao/dss-simulation-scripts. (also ref: Kenton’s script: https://github.com/makerdao/cage-keeper)

PR #264, opened 2022-03-22
Housekeeping

julienmartinlevrai:
* remove warnings
* update titles
* update copyrights
* update license notices
* update pragmas

Note that these warnings are a security issue for the spellcrafting process.

### DIFF
--- a/src/abaci.sol
+++ b/src/abaci.sol
@@ -17,7 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity 0.6.12;
+pragma solidity ^0.6.12;
 
 interface Abacus {
     // 1st arg: initial price               [ray]
--- a/src/abaci.sol
+++ b/src/abaci.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-// Copyright (C) 2020 Maker Ecosystem Growth Holdings, INC.
+/// abaci.sol -- price decrease functions for auctions
+
+// Copyright (C) 2020-2022 Dai Foundation
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published
@@ -15,7 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity >=0.6.12;
+pragma solidity 0.6.12;
 
 interface Abacus {
     // 1st arg: initial price               [ray]
--- a/src/end.sol
+++ b/src/end.sol
@@ -4,6 +4,7 @@
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
 // Copyright (C) 2018 Lev Livnev <lev@liv.nev.org.uk>
+// Copyright (C) 2020-2021 Maker Ecosystem Growth Holdings, INC.
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
--- a/src/end.sol
+++ b/src/end.sol
@@ -4,7 +4,7 @@
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
 // Copyright (C) 2018 Lev Livnev <lev@liv.nev.org.uk>
-// Copyright (C) 2020-2021 Maker Ecosystem Growth Holdings, INC.
+// Copyright (C) 2020-2022 Dai Foundation
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
@@ -19,7 +19,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity >=0.6.12;
+pragma solidity 0.6.12;
 
 interface VatLike {
     function dai(address) external view returns (uint256);
--- a/src/end.sol
+++ b/src/end.sol
@@ -19,7 +19,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity 0.6.12;
+pragma solidity ^0.6.12;
 
 interface VatLike {
     function dai(address) external view returns (uint256);
--- a/src/end.sol
+++ b/src/end.sol
@@ -4,7 +4,7 @@
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
 // Copyright (C) 2018 Lev Livnev <lev@liv.nev.org.uk>
-// Copyright (C) 2020-2021 Maker Ecosystem Growth Holdings, INC.
+// Copyright (C) 2020-2021 Dai Foundation
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
--- a/src/end.sol
+++ b/src/end.sol
@@ -4,7 +4,6 @@
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
 // Copyright (C) 2018 Lev Livnev <lev@liv.nev.org.uk>
-// Copyright (C) 2020-2022 Dai Foundation
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -17,7 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity 0.6.12;
+pragma solidity ^0.6.12;
 
 // FIXME: This contract was altered compared to the production version.
 // It doesn't use LibNote anymore.
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -3,7 +3,6 @@
 /// flop.sol -- Debt auction
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
-// Copyright (C) 2022 Dai Foundation
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
--- a/src/flop.sol
+++ b/src/flop.sol
@@ -3,6 +3,7 @@
 /// flop.sol -- Debt auction
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
+// Copyright (C) 2022 Dai Foundation
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU Affero General Public License as published by
@@ -17,7 +18,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity >=0.5.12;
+pragma solidity 0.6.12;
 
 // FIXME: This contract was altered compared to the production version.
 // It doesn't use LibNote anymore.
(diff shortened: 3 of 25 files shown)


gbalabasquer:
Although I haven't reviewed them yet, I think I'm fine with the changes.
The only one I'm not 100% sure is the internal functions naming. I'd be pretty sure changing those names wouldn't change the bytecode, but can we confirm that? If the bytecode changes, then I'd say not to do it.

kmbarry1:
I don't see any issues with these changes. Using version-locked pragmas is probably a practice we should standardize on more widely. To Gonza's concern...it seems quite unlikely, but it wouldn't hurt to check I guess.

julienmartinlevrai:
I am going to check if the bytecode changes. However, I also don’t think it matters much. If the concern is about keeping the same bytecode as the deployed contracts, remember that it is not the case anymore as we removed the lognote events.

julienmartinlevrai:
The bytecode does change @gbalabasquer

kmbarry1:
> The bytecode does change @gbalabasquer

Does it change only the hash at the end, or does the functional portion change too? Note that there's code in the tests of `spells-mainnet` that shows how to separate the hash from the bytecode.

If only the hash is changing, I do not believe that necessitates maintaining another branch.

kmbarry1 (dismissed):
LGTM

gbalabasquer:
> @gbalabasquer the only contracts that change are `vat`, `vow` and `pot`.

Ah yes you are right, why the newer contracts do not throw warnings about the `add` and `sub` functions?

julienmartinlevrai:
> > @gbalabasquer the only contracts that change are `vat`, `vow` and `pot`.
> 
> Ah yes you are right, why the newer contracts do not throw warnings about the `add` and `sub` functions?

Because it happens only when Assembly is used in the contract, because it interferes with the Yul functions of the same name.

gbalabasquer:
> > > @gbalabasquer the only contracts that change are `vat`, `vow` and `pot`.
> > 
> > 
> > Ah yes you are right, why the newer contracts do not throw warnings about the `add` and `sub` functions?
> 
> Because it happens only when Assembly is used in the contract, because Assembly has native functions with the same name.

Ohh ok, didn't know that was only in that case.
Now I want to change it everywhere, lol

src/join.sol
```diff
@@ -3,6 +3,7 @@
 /// join.sol -- Basic token adapters
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
+// Copyright (C) 2022 Dai Foundation
```
gbalabasquer, line 6: Same here.

src/jug.sol
```diff
@@ -3,6 +3,7 @@
 /// jug.sol -- Dai Lending Rate
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
+// Copyright (C) 2022 Dai Foundation
```
gbalabasquer, line 6: Same

src/test/abaci.t.sol
```diff
@@ -1,6 +1,23 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-pragma solidity >=0.6.12;
+// abaci.t.sol -- tests for abaci.sol
+
+// Copyright (C) 2018-2022 Dai Foundation
```
gbalabasquer, line 5: I think this should be:
```suggestion
// Copyright (C) 2020-2022 Dai Foundation
```

src/test/clip.t.sol
```diff
@@ -1,6 +1,23 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-pragma solidity >=0.6.12;
+// clip.t.sol -- tests for clip.sol
+
+// Copyright (C) 2018-2022 Dai Foundation
```
gbalabasquer, line 5: Same here:
```suggestion
// Copyright (C) 2020-2022 Dai Foundation
```

src/test/dog.t.sol
```diff
@@ -1,6 +1,23 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-pragma solidity >=0.6.12;
+// dog.t.sol -- tests for dog.sol
+
+// Copyright (C) 2018-2022 Dai Foundation
```
gbalabasquer, line 5:
```suggestion
// Copyright (C) 2020-2022 Dai Foundation
```

src/flop.sol
```diff
@@ -3,6 +3,7 @@
 /// flop.sol -- Debt auction
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
+// Copyright (C) 2022 Dai Foundation
```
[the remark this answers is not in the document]
julienmartinlevrai: AFAIU, the Dai Foundation already has the copyright of this code. I am just making it explicit.

src/vow.sol
```diff
@@ -3,6 +3,7 @@
 /// vow.sol -- Dai settlement module
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
+// Copyright (C) 2022 Dai Foundation
```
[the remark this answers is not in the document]
julienmartinlevrai: Check out my other comment, I think the Dai Foundation has the copyright anyways

src/spot.sol
```diff
@@ -2,6 +2,8 @@
 
 /// spot.sol -- Spotter
 
+// Copyright (C) 2018-2022 Dai Foundation
```
julienmartinlevrai, line 5:
```suggestion
// Copyright (C) 2020-2022 Dai Foundation
```

src/flop.sol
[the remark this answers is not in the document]
gbalabasquer: Hmmm we should double check, but I don't think the code was transferred per se like it happen with the ones written by the Maker Foundation.
So I think we only add the pragma if we enter modifications to the code, like a second writer.

src/test/flap.t.sol
```diff
@@ -1,6 +1,23 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-pragma solidity >=0.5.12;
+// flap.t.sol -- tests for flap.sol
+
+// Copyright (C) 2018-2022 Dai Foundation
```
julienmartinlevrai, line 5:
```suggestion
// Copyright (C) 2020-2022 Dai Foundation
```

src/test/flop.t.sol
```diff
@@ -1,6 +1,23 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-pragma solidity >=0.5.12;
+// flop.t.sol -- tests for flop.sol
+
+// Copyright (C) 2018-2022 Dai Foundation
```
julienmartinlevrai, line 5:
```suggestion
// Copyright (C) 2020-2022 Dai Foundation
```

src/flop.sol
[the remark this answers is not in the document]
gbalabasquer: As otherwise we should have directly replace them to the Dai Foundation, like we did with MEGH.

src/abaci.sol
```diff
@@ -15,7 +17,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity >=0.6.12;
```
brianmcmichael, line 18: I may have missed a discussion around this, but locking these may break any repos that accept this as a dependency that are using a later (but compatible) version. I don't think this should go do master.

gbalabasquer: I thought about the same thing, and was about to mention it. But solidity `0.6.12` is the last one and haven't been a new one for a long while. So at this stage seems very unlikely a new one happens, then is the same in practice.

julienmartinlevrai: @naszam made me aware of this best practice advice by ConsenSys: https://consensys.github.io/smart-contract-best-practices/development-recommendations/solidity-specific/locking-pragmas/

Initially I was favor of something like `~0.6.12` which allows for a patch bump, but the problem is that it also compiles with nightly builds.

naszam: yeah, this was discussed in https://github.com/makerdao/MIP21-RWA-Example/pull/7 as well.

src/flop.sol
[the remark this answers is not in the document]
julienmartinlevrai: After a talk with Soren, you are right. Updated.

src/abaci.sol
gbalabasquer: Well let's not compare MIP21 with this repo that they have nothing to do between.
This repo is imported in several other ones. Locking to a version might provoke a nightmare, if it wasn't that we are in the latest version and we do not think there will be a new one I would be in opposition to it.
Best practices are subjective, they are also relative to the situation. So sometimes what Slither or Consensys says doesn't really matter.
Also IMO, this one kind of lacks of real value, the pragma for example doesn't ensure you are compiling with the specific intended flags.
If you want to make sure to be compiling exactly as was tested then build a script, use a docker image or nix procedure to make sure it is being done as corresponds.

src/end.sol
```diff
@@ -4,7 +4,6 @@
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
 // Copyright (C) 2018 Lev Livnev <lev@liv.nev.org.uk>
-// Copyright (C) 2020-2021 Maker Ecosystem Growth Holdings, INC.
```
gbalabasquer, line 7: Please add this line back with the Dai Foundation. We introduced changes to the end during those years, that's why the copyright was added.

src/test/flap.t.sol
```diff
@@ -1,6 +1,23 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-pragma solidity >=0.5.12;
+// flap.t.sol -- tests for flap.sol
+
+// Copyright (C) 2022 Dai Foundation
```
gbalabasquer, line 5: Hmm the copyright of this file shouldn't be in the Dai Foundation

gbalabasquer: Sorry nvm.

src/end.sol
julienmartinlevrai: done

src/abaci.sol
naszam: The discussion I was referring to is https://github.com/makerdao/MIP21-RWA-Example/pull/7#discussion_r809173471, where a locked pragma `0.5.12` was replaced by `~0.6.12`.
Regarding pragma best practises, as stated in the [SWC-13](https://swcregistry.io/docs/SWC-103):
> Pragma statements can be allowed to float when a contract is intended for consumption by other developers, as in the case with contracts in a library or EthPM package. Otherwise, the developer would need to manually update the pragma in order to compile locally.

So, in case of contract used as dependency in other repos, I think we should keep the floating pragma (e.g. `>=0.6.12`)

julienmartinlevrai: Fair enough. How about we use the [caret range](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004) `^0.6.12` (which I just realized is the same as a tilde range) so that only a patch bump is allowed

src/end.sol
```diff
@@ -4,6 +4,7 @@
 
 // Copyright (C) 2018 Rain <rainbreak@riseup.net>
 // Copyright (C) 2018 Lev Livnev <lev@liv.nev.org.uk>
+// Copyright (C) 2020-2021 Maker Ecosystem Growth Holdings, INC.
```
gbalabasquer, line 7: Well should still be replaced by the Dai Foundation.
So whatever was under MEGH should be moved to Dai Foundation
```suggestion
// Copyright (C) 2020-2021 Dai Foundation
```

src/abaci.sol
brianmcmichael: There may not be active development on 0.6.x right now but the discovery of a security vulnerability sometimes causes devs to go back and patch earlier versions.

julienmartinlevrai: done

brianmcmichael (approved):

gbalabasquer (approved):
LGTM

PR #267, opened 2022-06-07
Revert on end.cage(ilk) if there is not Art for that ilk

gbalabasquer:
So the change didn't result as simple as it seemed.
Adding the revert condition is a bit problematic for `skip` and `snip` if all the debt of that ilk is in running auctions (that was exactly the case of the tests).

Without a `tag` the auctions can't be yanked. This case is very improbably in practice as for a specific `ilk` would be super rare there isn't at least one open CDP which passes the new `Art` `require` condition.
However if that happened, it would mean that auctions can't be yanked nor collateral caged leaving it like in a trap (well still can finish through the normal auction path).

So the correct thing to do would be removing the `require` of `tag > 0` in `skip` and `snip`. I still need to do a better review but seems wouldn't be a problem. If `cage(ilk)` is called after some `skip` or `snip` nothing bad should happen as the `Art` itself will be overwritten taking into account again the auction that was previously yanked. And `tag` doesn't really seem to be necessary to run `skip` or `snip`.

I'd like to hear some opinions thought before thinking on deeper test cases. Do we feel we can continue this path or should we reconsider it?

### DIFF
--- a/src/end.sol
+++ b/src/end.sol
@@ -347,7 +347,9 @@ contract End {
     function cage(bytes32 ilk) external {
         require(live == 0, "End/still-live");
         require(tag[ilk] == 0, "End/tag-ilk-already-defined");
-        (Art[ilk],,,,) = vat.ilks(ilk);
+        (uint256 Art_,,,,) = vat.ilks(ilk);
+        require(Art_ > 0, "End/Art-is-zero");
+        Art[ilk] = Art_;
         (PipLike pip,) = spot.ilks(ilk);
         // par is a ray, pip returns a wad
         tag[ilk] = wdiv(spot.par(), uint256(pip.read()));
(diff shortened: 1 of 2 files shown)


hexonaut:
Removing these requires makes me a bit nervous. We have a solution with the oracle that albeit is a bit messier, but it's easier to reason that it is safe. I'm leaning more towards the oracle hack solution now.

gbalabasquer:
> Removing these requires makes me a bit nervous. We have a solution with the oracle that albeit is a bit messier, but it's easier to reason that it is safe. I'm leaning more towards the oracle hack solution now.

Yes, having to remove these two `require` changed a bit the panorama.
I'm also feeling that we might go with another route.

iamchrissmith:
I agree with @hexonaut here, the requires makes me lean towards the oracle solution

